PR #3, opened 2020-06-24
Add appengine-web.xml and fix artifact ID

avagarcia0:
Making these changes allowed me to run the project locally.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,0 +1,15 @@
+{
+  "parserOptions": {
+    "ecmaVersion": 2017
+  },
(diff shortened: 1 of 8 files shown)


emilyvera:
Why does google cloud build fail? What does that even mean

acheng1:
> Why does google cloud build fail? What does that even mean

I enabled that when I linked the GCP project but I just disabled it. If you push an empty commit `git commit --amend`, it should no longer run.

avagarcia0:
That fixed it, thanks!

PR #2, opened 2020-06-24
Add initial pom file and lint checks to project

acheng1:
You can also run these commands locally

`make pretty to format code`
`make validate to validate the code`

PR #4, opened 2020-06-24
Add initial pom file and lint checks to project

acheng1:
You can also run these commands locally

`make pretty to format code`
`make validate to validate the code`

This is the same as https://github.com/googleinterns/step12-2020/pull/2 but I had to update the email address I used to commit (for the CLA validation).


acheng1:
Same as https://github.com/googleinterns/step12-2020/pull/2 but I had to fix the email address I used to commit.

PR #7, opened 2020-06-26
Add servlet to upload receipt image using Blobstore.

basaam0:


src/main/java/servlets/UploadReceiptServlet.java
```diff
@@ -0,0 +1,174 @@
...
+    receipt.setProperty("label", label);
+    receipt.setProperty("price", price);
+    receipt.setProperty("store", store);
+    receipt.setProperty("rawText", rawText);
+
+    return Optional.of(receipt);
+  }
+
+  /**
+   * Returns a URL that points to the uploaded file, or null if the user didn't upload a JPEG file.
```
avagarcia0, line 128: You appear to be returning an empty Optional rather than null if a JPEG file wasn't uploaded.

emilyvera: Maybe we can also display a popup to the user that they didn't select a file

basaam0: Added an error message.

src/main/java/servlets/UploadReceiptServlet.java
```diff
@@ -0,0 +1,176 @@
...
+  /**
+   * Creates a URL that uploads the receipt image to Blobstore when the user submits the upload
+   * form. After Blobstore handles the parsing, storing, and hosting of the image, the form
+   * data and a URL where the image can be accessed is forwarded to this servlet in a POST
+   * request.
+   */
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    BlobstoreService blobstoreService = BlobstoreServiceFactory.getBlobstoreService();
+    String uploadUrl = blobstoreService.createUploadUrl("/upload-receipt");
```
acheng1, line 62: Do we want to set UploadOptions (e.g., https://cloud.google.com/appengine/docs/standard/java/javadoc/com/google/appengine/api/blobstore/UploadOptions.html#maxUploadSizeBytesPerBlob-long-)

src/main/webapp/js/upload.js
```diff
@@ -33,3 +42,15 @@ async function fetchBlobstoreUrlAndShowForm() {
   // The form is hidden by default.
   uploadForm.classList.remove('hidden');
 }
+
+/**
+ * Displays an error message if the user did not upload a JPEG file.
```
acheng1, line 47: Could you clarify how this works? Why is the error in the query string parameters?

src/main/java/servlets/UploadReceiptServlet.java
```diff
@@ -0,0 +1,176 @@
...
+    receipt.setProperty("rawText", rawText);
+
+    return Optional.of(receipt);
+  }
+
+  /**
+   * Returns a URL that points to the uploaded file, or an empty Optional if the user didn't upload
+   * a JPEG file.
+   */
+  private Optional<BlobKey> getUploadedBlobKey(
```
acheng1, line 132: Really nice, love that this follows go/tott/455!

src/main/java/servlets/UploadReceiptServlet.java
basaam0: Added UploadOptions for a 5 MB maximum

src/main/webapp/js/upload.js
basaam0: Initially, if the user submitted the form with a non-JPEG file, the servlet would redirect back to the upload page with an error flag in the URL parameter indicating to show a popup. 

I have now updated the logic to fetch the servlet response and redirect to the home page from the client-side. I added restrictions on the file type accepted on the front-end. If those are bypassed and the form is submitted with a non-JPEG file, the servlet will respond with an error code and a popup message will be shown.

src/main/webapp/css/upload.css
```diff
@@ -1,7 +1,7 @@
 /* Form is hidden by default. */
-.hidden {
+/* .hidden {
   display: none;
-}
+} */
```
acheng1, line 4: Is this no longer needed?

src/main/java/servlets/UploadReceiptServlet.java
```diff
@@ -0,0 +1,176 @@
...
+import javax.servlet.http.HttpServletResponse;
+
+/**
+ * Servlet with a GET handler that creates a URL that uploads a receipt image to Blobstore and
+ * a POST handler that extracts data from the image and inserts it into Datastore.
+ */
+@WebServlet("/upload-receipt")
+public class UploadReceiptServlet extends HttpServlet {
+  // Max upload size of 5 MB.
+  private static final long MAX_UPLOAD_SIZE_BYTES = 5242880;
```
acheng1, line 48: really minor nit, but I prefer to use an expression here like 5 * 1024 * 1024 just to avoid any typos.

src/main/webapp/css/upload.css
basaam0: Yes, now the form will be available immediately and the Blobstore upload URL will be created after it is submitted. I have removed the code.

PR #10, opened 2020-06-29
Add a servlet that uses the Vision API to serve the text of a given image

avagarcia0:


src/main/java/servlets/ReceiptAnalysisServlet.java
```diff
@@ -0,0 +1,81 @@
...
+    if (filePath == null) {
+      return;
+    }
+
+    response.setContentType("text/html");
+    response.getWriter().println(retrieveText(filePath));
+  }
+
+  /** Detects and retrieves text in the specified image. */
+  public static String retrieveText(String filePath) {
```
basaam0, line 55: Does this need to be public?

src/main/java/servlets/ReceiptAnalysisServlet.java
```diff
@@ -0,0 +1,81 @@
...
+      requests.add(request);
+
+      try (ImageAnnotatorClient client = ImageAnnotatorClient.create()) {
+        BatchAnnotateImagesResponse batchResponse = client.batchAnnotateImages(requests);
+        AnnotateImageResponse response = batchResponse.getResponsesList().get(0);
+        EntityAnnotation annotation = response.getTextAnnotationsList().get(0);
+
+        description = annotation.getDescription();
+      }
+    } catch (IOException e) {
```
emilyvera, line 76: Do we want to leave this catch statement empty?

src/main/java/servlets/ReceiptAnalysisServlet.java
```diff
@@ -0,0 +1,81 @@
...
+    }
```
basaam0, line 77: From the Google Java style guide:

>  [I]t is very rarely correct to do nothing in response to a caught exception.
> When it truly is appropriate to take no action whatsoever in a catch block, the reason this is justified is explained in a comment.

https://google.github.io/styleguide/javaguide.html#s6.2-caught-exceptions

src/main/java/servlets/ReceiptAnalysisServlet.java
emilyvera, line 55: This is minor, but I previously received feedback that helper functions only used in this class should be private instead of public.

src/main/java/servlets/ReceiptAnalysisServlet.java
avagarcia0: Should I make an AnalysisResults object that includes a field for indicating whether an error occurred, and use Gson to send it?

basaam0: If the object only has one field, it may be better to use the HttpServletResponse.sendError() to respond with a status code and an error message.

https://tomcat.apache.org/tomcat-5.5-doc/servletapi/javax/servlet/http/HttpServletResponse.html#sendError(int,%20java.lang.String)

src/main/java/data/AnalysisResults.java
```diff
@@ -0,0 +1,26 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.sps.data;
+
+/**
+ * Object for holding the analysis results served by ReceiptAnalysisServlet.
+ */
+public class AnalysisResults {
+  String rawText;
```
basaam0, line 21: Should be declared private for encapsulation.

src/main/java/servlets/ReceiptAnalysisServlet.java
```diff
@@ -0,0 +1,105 @@
...
+
+    try {
+      results = new AnalysisResults(retrieveText(imgBytes));
+    } catch (IOException e) {
+      response.sendError(HttpServletResponse.SC_INTERNAL_SERVER_ERROR, "Failed to retrieve text.");
+      return;
+    }
+
+    Gson gson = new Gson();
+    response.setContentType("text/html;");
```
basaam0, line 69: The content type is now application/json

src/main/java/servlets/ReceiptAnalysisServlet.java
```diff
@@ -0,0 +1,105 @@
...
+    response.getWriter().println(gson.toJson(results));
+  }
+
+  /** Reads the image bytes from the specified path. */
+  private ByteString readImageBytes(String filePath) throws IOException {
+    ByteString imgBytes;
+
+    try (InputStream fileInputStream = new FileInputStream(filePath)) {
+      imgBytes = ByteString.readFrom(fileInputStream);
+    }
```
basaam0, line 79: Is this try block or the one on line 95 necessary?

src/main/java/servlets/ReceiptAnalysisServlet.java
avagarcia0: Huh, I missed that during the json tutorial.

src/main/java/data/AnalysisResults.java
avagarcia0: This class can't be used if it's private.

src/main/java/servlets/ReceiptAnalysisServlet.java
avagarcia0: Yes, they are try-with-resources statements.

src/main/java/servlets/ReceiptAnalysisServlet.java
```diff
@@ -0,0 +1,105 @@
...
+      return;
+    }
+
+    Gson gson = new Gson();
+    response.setContentType("text/html;");
+    response.getWriter().println(gson.toJson(results));
+  }
+
+  /** Reads the image bytes from the specified path. */
+  private ByteString readImageBytes(String filePath) throws IOException {
```
basaam0, line 74: Will this work for file paths like "/serve-image?blob-key=Wyd8EBnLuOdjVsZCUmtl_w"? The receipt images are served by the ServeImageServlet and can be accessed through a GET request with the blob key as a parameter.

src/main/java/data/AnalysisResults.java
basaam0: Sorry, I meant the rawText field on line 21 should be marked private.

src/main/java/servlets/ReceiptAnalysisServlet.java
avagarcia0: This should now work for the absolute form of that URL, though you might have to use a URL escaper: https://guava.dev/releases/19.0/api/docs/com/google/common/net/UrlEscapers.html

src/main/java/data/AnalysisResults.java
```diff
@@ -0,0 +1,26 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.sps.data;
+
+/**
+ * Object for holding the analysis results served by ReceiptAnalysisServlet.
+ */
+public class AnalysisResults {
+  private String rawText;
```
acheng1, line 21: Always mark an instance variable (field) final if it is never reassigned (e.g., private final String rawText;). See Effective Java Item 17 "Minimize mutability".

src/main/java/servlets/ReceiptAnalysis.java
```diff
@@ -0,0 +1,77 @@
...
+    return imgBytes;
+  }
+
+  /** Detects and retrieves text in the provided image. */
+  private static AnalysisResults retrieveText(ByteString imgBytes) throws IOException {
+    List<AnnotateImageRequest> requests = new ArrayList<>();
+    String description = "";
+
+    Image img = Image.newBuilder().setContent(imgBytes).build();
+    Feature feat = Feature.newBuilder().setType(Feature.Type.TEXT_DETECTION).build();
```
acheng1, line 62: nit: rename to feature instead of "feat".

General advice on naming: Use the shortest possible name, but not at the cost of clarity. Avoid acronyms, abbreviations, and in-team jargon.

src/main/java/servlets/ReceiptAnalysis.java
```diff
@@ -0,0 +1,77 @@
...
+
+    return imgBytes;
+  }
+
+  /** Detects and retrieves text in the provided image. */
+  private static AnalysisResults retrieveText(ByteString imgBytes) throws IOException {
+    List<AnnotateImageRequest> requests = new ArrayList<>();
+    String description = "";
+
+    Image img = Image.newBuilder().setContent(imgBytes).build();
```
acheng1, line 61: nit: similar comment, avoid abbreviating and rename to "image".

src/main/java/servlets/ReceiptAnalysis.java
```diff
@@ -0,0 +1,77 @@
...
+    Feature feat = Feature.newBuilder().setType(Feature.Type.TEXT_DETECTION).build();
+    AnnotateImageRequest request =
+        AnnotateImageRequest.newBuilder().addFeatures(feat).setImage(img).build();
+    requests.add(request);
+
+    try (ImageAnnotatorClient client = ImageAnnotatorClient.create()) {
+      BatchAnnotateImagesResponse batchResponse = client.batchAnnotateImages(requests);
+      AnnotateImageResponse response = batchResponse.getResponsesList().get(0);
```
acheng1, line 69: Is it possible for the response list to be empty? Should we check if the response.hasError()?

src/main/java/servlets/ReceiptAnalysis.java
```diff
@@ -0,0 +1,77 @@
...
+      EntityAnnotation annotation = response.getTextAnnotationsList().get(0);
```
acheng1, line 70: Is it also possible for `getTextAnnotationsList()` to return an empty list?

acheng1, line 70: nit: maybe add a comment about why we only get the first element?

src/main/java/servlets/ReceiptAnalysis.java
bryyeh, line 61: nit: I agree with Tony. I would rename this to "image".

src/main/java/servlets/ReceiptAnalysis.java
```diff
@@ -0,0 +1,77 @@
...
+  /** Detects and retrieves text in the provided image. */
+  private static AnalysisResults retrieveText(ByteString imgBytes) throws IOException {
+    List<AnnotateImageRequest> requests = new ArrayList<>();
+    String description = "";
+
+    Image img = Image.newBuilder().setContent(imgBytes).build();
+    Feature feat = Feature.newBuilder().setType(Feature.Type.TEXT_DETECTION).build();
+    AnnotateImageRequest request =
+        AnnotateImageRequest.newBuilder().addFeatures(feat).setImage(img).build();
+    requests.add(request);
```
bryyeh, line 65: Instead of initializing the list on L58 and then adding the request here, I would do both on this line. This way the logic isn't in two places. I also recommend you use an ImmutableList when possible. So this line could be 
`ImmutableList<AnnotateImageRequest> requests = ImmutableList.of(request)`

src/main/java/servlets/ReceiptAnalysis.java
bryyeh: I like to use Iterables.getOnlyElement() for returning the single item in a list.

https://guava.dev/releases/21.0/api/docs/com/google/common/collect/Iterables.html

src/main/java/servlets/ReceiptAnalysis.java
avagarcia0: We probably should handle both those cases, though I'm not sure what the best way to do that is.

avagarcia0: I just saw that Basaam added a ReceiptAnalysisException in #12. Would it make sense to throw that in the event that we detect one of these issues?

src/main/java/servlets/ReceiptAnalysis.java
avagarcia0: @bryyeh I'm just getting the first item in the list, not the only item in it.

src/main/java/servlets/ReceiptAnalysis.java
acheng1: That seems reasonable to me.

src/main/java/servlets/ReceiptAnalysis.java
bryyeh: Oh, what are the other items in the list and why are you only getting the first one?

src/main/java/servlets/ReceiptAnalysis.java
bryyeh: While Basaam's PR is being reviewed, maybe we can add a TODO here so this code can be submitted.

src/main/java/servlets/ReceiptAnalysis.java
avagarcia0: The other items are the individual words from the raw text.

src/main/java/servlets/ReceiptAnalysis.java
avagarcia0: Alright, I added one.

src/main/java/servlets/ReceiptAnalysis.java
```diff
@@ -64,6 +64,7 @@ private static AnalysisResults retrieveText(ByteString imageBytes) throws IOExce
     ImmutableList<AnnotateImageRequest> requests = ImmutableList.of(request);
 
     try (ImageAnnotatorClient client = ImageAnnotatorClient.create()) {
+      // TODO: Throw custom exception if response has an error or is missing
```
bryyeh, line 67: Can you also put the pull request you're waiting for in this TODO. This way someone reading through the code knows why it hasn't been done yet.

PR #11, opened 2020-06-30
Receipt search servlet

emilyvera:
- Add some dependencies to pom.xml
- Add Receipt.java class to represent a receipt object
- Search by label working with call to datastore (I put some test objects in)
- Dynamic UI that changes based on how many receipts are found/displayed

You can test out the retrieval feature at https://capstone-receipt-step-2020.uc.r.appspot.com/.
Notes:
- UI looks a bit better on mobile
- Type in "test" or "hello" in the search bar to see two different queries 
- Results may take a few seconds to load on the first try (we should discuss how to make this faster)
- If not results were found, a simple error message is shown to the user
- Buttons like log out/add a receipt/view/edit/delete do nothing for now
- The date-range drop down works but this isn't taken into account in the search query yet


src/main/java/data/Receipt.java
```diff
@@ -0,0 +1,46 @@
...
+import com.google.appengine.api.blobstore.BlobKey;
+import java.util.Set;
+
+/* Class to represent a receipt and its properties.*/
+public class Receipt {
+  private long id;
+  private long userId;
+  private long timestamp;
+  private BlobKey blobkey;
+  private String imageURL;
```
basaam0, line 26: The Java style guide defines camel case disregarding the casing of the original word, so in camel case this would be imageUrl.

https://google.github.io/styleguide/javaguide.html#s5.3-camel-case

src/main/java/data/Receipt.java
```diff
@@ -0,0 +1,46 @@
...
+      double price, String store, String label, Set<String> categories, String rawText) {
+    this.id = id;
+    this.userId = userId;
+    this.timestamp = timestamp;
+    this.blobkey = blobkey;
+    this.imageURL = imageURL;
+    this.price = price;
+    this.store = store;
+    this.label = label;
+    this.categories = categories;
```
basaam0, line 43: It might be better to make a deep copy using addAll() to prevent leaking of access.

emilyvera: I was debating on making this a shallow or deep copy, but you're right

src/main/webapp/js/index.js
```diff
@@ -12,6 +12,66 @@
...
+  receiptsDisplayed.innerHTML = '';
+
+  receipts.forEach((receipt) => {
+    receiptsDisplayed.innerHTML += createReceiptCardElement(receipt);
+  });
+}
+
+/**
+ * Returns an HTML string that represents a receipt card.
+ * This card displays trasaction date, store name, trasaction total, categories,
```
basaam0, line 32: Typo for "transaction"

src/main/java/servlets/SearchServlet.java
```diff
@@ -0,0 +1,80 @@
...
+      double price = (double) entity.getProperty("price");
+      String store = (String) entity.getProperty("store");
+      String label = (String) entity.getProperty("label");
+      Set<String> categories = new HashSet<String>((ArrayList) entity.getProperty("categories"));
+      String rawText = (String) entity.getProperty("rawText");
+
+      Receipt receipt = new Receipt(
+          id, userId, timestamp, blobkey, imageURL, price, store, label, categories, rawText);
+      receipts.add(receipt);
+    }
```
basaam0, line 77: I have received feedback that this would be a good place to use Streams, which are more declarative.
https://docs.oracle.com/javase/8/docs/api/?java/util/stream/Stream.html

The code would be something like:
```
Stream<Entity> entities = ...
List<Receipt> receipts = entities.map(entity -> {
  ...
  return new Receipt(...);
}).collect(Collectors.toList());
```

pom.xml
```diff
@@ -24,13 +24,27 @@
...
+
     <dependency>
       <groupId>com.google.appengine</groupId>
       <artifactId>appengine-api-1.0-sdk</artifactId>
       <version>1.9.59</version>
     </dependency>
   </dependencies>
 
+  
+
```
acheng1, line 47: nit: remove extra newlines

pom.xml
```diff
@@ -24,13 +24,25 @@
...
+    </dependency>
+
     <dependency>
       <groupId>com.google.appengine</groupId>
       <artifactId>appengine-api-1.0-sdk</artifactId>
       <version>1.9.59</version>
     </dependency>
   </dependencies>
-
+
```
acheng1, line 45: nitnitnit: are there extra spaces here?

src/main/java/data/Receipt.java
```diff
@@ -0,0 +1,47 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.sps.data;
+
+import com.google.appengine.api.blobstore.BlobKey;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+/* Class to represent a receipt and its properties.*/
```
acheng1, line 21: nitnitnit: add space after the period

src/main/java/data/Receipt.java
```diff
@@ -0,0 +1,47 @@
...
+public class Receipt {
+  private long id;
+  private long userId;
+  private long timestamp;
+  private BlobKey blobKey;
+  private String imageUrl;
+  private double price;
+  private String store;
+  private String label;
+  private Set<String> categories;
```
acheng1, line 31: Prefer `ImmutableSet` (see https://github.com/google/guava/wiki/ImmutableCollectionsExplained for advantages) if you don't plan to modify `categories` later.

In the constructor you can then do the following
```
this.categories = ImmutableSet.copyOf(categories);
```

src/main/java/data/Receipt.java
```diff
@@ -0,0 +1,47 @@
...
+
+package com.google.sps.data;
+
+import com.google.appengine.api.blobstore.BlobKey;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+/* Class to represent a receipt and its properties.*/
+public class Receipt {
+  private long id;
```
acheng1, line 23: Always mark an instance variable (field) final if it is never reassigned (e.g., `private final long id;`). See Effective Java Item 17 "Minimize mutability".

src/main/java/data/Receipt.java
```diff
@@ -0,0 +1,47 @@
...
+  private long timestamp;
+  private BlobKey blobKey;
+  private String imageUrl;
+  private double price;
+  private String store;
+  private String label;
+  private Set<String> categories;
+  private String rawText;
+
+  public Receipt(long id, long userId, long timestamp, BlobKey blobKey, String imageUrl,
```
acheng1, line 34: This is optional so no need to change your current implementation but when there are a lot of parameters we tend to use the builder pattern (see Effective Java Item 2). 

```
Receipt receipt = new Receipt.Builder().id(1).userId(2).timestamp(1).blobKey("abc").imageUrl("http://...").build();
```

src/main/java/servlets/SearchServlet.java
```diff
@@ -0,0 +1,85 @@
...
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    String desiredLabel = request.getParameter("label");
+    List<Receipt> receipts = getReceiptsWithMatchingLabel(desiredLabel);
+
+    Gson gson = new Gson();
+    response.setContentType("application/json;");
+    response.getWriter().println(gson.toJson(receipts));
+  }
+
+  /* Return a list of receipts from datastore with the same label as desiredLabel.*/
```
acheng1, line 55: nitnitnit: add space after period.

src/main/java/servlets/SearchServlet.java
```diff
@@ -0,0 +1,85 @@
...
+  private List<Receipt> getReceiptsWithMatchingLabel(String desiredLabel) {
```
acheng1, line 56: Similar as mentioned above, prefer immutable collections. Here, `ImmutableList` instead of `List`.

src/main/java/servlets/SearchServlet.java
```diff
@@ -0,0 +1,85 @@
...
+    Filter matchingLabels = new FilterPredicate("label", FilterOperator.EQUAL, desiredLabel);
+    Query query = new Query("Receipt");
+    query.setFilter(matchingLabels);
+
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+
+    // Convert matching receipt entities to receipt objects and add to return list.
+    Spliterator<Entity> spliterator = datastore.prepare(query).asIterable().spliterator();
+    Stream<Entity> entities = StreamSupport.stream(spliterator, false);
+    return entities
```
acheng1, line 67: I think this could also be written as the following:

```
datastore.prepare(query).asList(FetchOptions.Builder.withDefaults()).stream().map(...)
```

src/main/webapp/js/index.js
```diff
@@ -12,6 +12,79 @@
...
+
+
+/**
+ * Returns an HTML string that represents a receipt card.
+ * This card displays transaction date, store name, trasaction total,
+ * categories, receipt photo, and view/edit/delete buttons.
+ */
+function createReceiptCardElement(receipt) {
+  const categories = Array.from(receipt.categories);
+  const formattedHTML = '<div class="col-md-6">' +
```
acheng1, line 50: I might try to put this in the html file itself using the `template` tag (See https://www.w3schools.com/tags/tag_template.asp) and use `document.getElementById()` to fill in the template.

From https://google.github.io/styleguide/htmlcssguide.html#Separation_of_Concerns :

Separate structure from presentation from behavior.
Strictly keep structure (markup), presentation (styling), and behavior (scripting) apart, and try to keep the interaction between the three to an absolute minimum.

src/main/java/data/Receipt.java
bryyeh: And if you really felt like trying something new, we like to use AutoValue to create boilerplate value classes.

https://github.com/google/auto/tree/master/value

src/main/webapp/js/index.js
emilyvera: I was looking for something like this but I guess I wasn't googling the right thing! Thanks

src/main/java/data/Receipt.java
```diff
@@ -0,0 +1,47 @@
...
+    this.id = id;
+    this.userId = userId;
+    this.timestamp = timestamp;
+    this.blobKey = blobKey;
+    this.imageUrl = imageUrl;
+    this.price = price;
+    this.store = store;
+    this.label = label;
+    this.categories = ImmutableSet.copyOf(categories);
+    ; // creates a deep copy
```
acheng1, line 44: ?

src/main/java/servlets/SearchServlet.java
```diff
@@ -0,0 +1,91 @@
...
+    Query query = new Query("Receipt");
+    query.setFilter(matchingLabels);
+
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+
+    // Convert matching receipt entities to receipt objects and add to return list.
+    return ImmutableList.copyOf(datastore.prepare(query)
+                                    .asList(FetchOptions.Builder.withDefaults())
+                                    .stream()
+                                    .map(entity -> { return createReceiptFromEntity(entity); })
```
acheng1, line 71: Normally, you should prefer method references to explicit lambdas, since the former usually result in shorter, cleaner code. For example:

`.map(this:: createReceiptFromEntity)`

src/main/webapp/index.html
```diff
@@ -1,6 +1,7 @@
 <!DOCTYPE html>
 <!-- Required meta tags -->
 <meta charset="utf-8" />
+<!-- <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no"> -->
```
acheng1, line 4: is this still needed?

src/main/webapp/js/index.js
```diff
@@ -12,8 +12,94 @@
...
+  displayReceipts(label, receipts);
+}
+
+/**
+ * Clear out existing receipts display.
+ *
+ * Note: 3 childNodes make up a receipt card.
+ * The first 3 nodes are the template receipt card and each subsequent
+ * 3 are clones of the template. The template must never be deleted or
+ * it'll be unusable in the future.
```
acheng1, line 32: Could the template be moved outside of receipts-display so you don't have to worry about accidentally deleting the template?

src/main/webapp/js/index.js
```diff
@@ -12,8 +12,94 @@
...
+ */
+function clearExistingReceipts() {
+  const receiptsDisplay = document.getElementById('receipts-display');
+  const numChildNodes = receiptsDisplay.childNodes.length;
```
acheng1, line 36: Maybe you could use the querySelector to get all the `<div class="col-md-6">` instead of all child nodes to delete each "receipt card" element at a time instead of each node at a time.

acheng1, line 36: Maybe you could use the querySelector to get all the `<div class="col-md-6">` instead of all child nodes to delete each "receipt card" element at a time instead of each node at a time.

src/main/java/servlets/SearchServlet.java
```diff
@@ -0,0 +1,91 @@
...
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    String desiredLabel = request.getParameter("label");
+    ImmutableList<Receipt> receipts = getReceiptsWithMatchingLabel(desiredLabel);
+
+    Gson gson = new Gson();
+    response.setContentType("application/json;");
+    response.getWriter().println(gson.toJson(receipts));
+  }
+
+  /* Return ImmutableList of receipts from datastore with the same label as desiredLabel. */
```
acheng1, line 58: nitnit: Write the phrase in the third person, as if there is an implied This method… before it. For example, "Returns ImmutableList of..."

src/main/webapp/js/index.js
```diff
@@ -12,8 +12,94 @@
...
+  templateScope.content.querySelector('#store-name').innerText = receipt.store;
+  templateScope.content.querySelector('#total').innerText =
+      'Total: $' + receipt.price;
+  templateScope.content.querySelector('#c1').innerText = categories[0];
+  templateScope.content.querySelector('#c2').innerText = categories[1];
+  templateScope.content.querySelector('#c3').innerText = categories[2];
+  templateScope.content.querySelector('img').src = receipt.imageUrl;
+
+  // Clone and attach receipt card to parent div.
+  const receiptCardClone = document.importNode(templateScope.content, true);
```
acheng1, line 82: Based on [1] importNode is used to import a node from another document (like an iframe). Although this will work, I think it's more correct to use `<node>.cloneNode()` within the same document.

[1] https://www.w3schools.com/jsref/met_document_importnode.asp

src/main/webapp/js/index.js
```diff
@@ -12,8 +12,94 @@
...
+  const templateScope = document.querySelector('#receipt-card-template');
+  templateScope.content.querySelector('#timestamp').innerText =
+      receipt.timestamp;
+  templateScope.content.querySelector('#store-name').innerText = receipt.store;
+  templateScope.content.querySelector('#total').innerText =
+      'Total: $' + receipt.price;
+  templateScope.content.querySelector('#c1').innerText = categories[0];
+  templateScope.content.querySelector('#c2').innerText = categories[1];
+  templateScope.content.querySelector('#c3').innerText = categories[2];
+  templateScope.content.querySelector('img').src = receipt.imageUrl;
```
acheng1, line 79: Why not clone the node first and make the changes to the cloned node instead of the template?

src/main/webapp/js/index.js
```diff
@@ -12,8 +12,94 @@
...
+ * @param {string} label User-entered label.
+ */
+function createErrorMessageElement(label) {
+  const div = document.createElement('div');
+  div.setAttribute('class', 'col-md-12 text-center error-message');
+  const h3 = document.createElement('h3');
+  h3.textContent = 'Sorry, no results found for "' + label +
+      '". Please try your search again or try a different query.';
+  div.appendChild(h3);
+  document.getElementById('receipts-display').appendChild(div);
```
acheng1, line 97: You could also probably use a template here as well.

src/main/java/servlets/SearchServlet.java
```diff
@@ -62,24 +65,27 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
-          String rawText = (String) entity.getProperty("rawText");
-          return new Receipt(
-              id, userId, timestamp, blobKey, imageUrl, price, store, label, categories, rawText);
-        })
-        .collect(Collectors.toList());
+    return ImmutableList.copyOf(datastore.prepare(query)
+                                    .asList(FetchOptions.Builder.withDefaults())
+                                    .stream()
+                                    .map(entity -> { return createReceiptFromEntity(entity); })
+                                    .collect(Collectors.toList()));
```
bryyeh, line 72: You can use ImmutableList.toImmuatbleList(), which returns a collector for immutable list. This way you don't need to do ImmutableList.copyOf()

src/main/java/servlets/SearchServlet.java
```diff
@@ -62,24 +65,27 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
+  }
+
+  /** Create and return a java Receipt object from a datastore Receipt entity. */
```
bryyeh, line 75: You can use `{@link Receipt}` to directly reference the Receipt object in javadocs. Same could be done with Entity.

src/main/webapp/js/index.js
```diff
@@ -12,81 +12,94 @@
...
-      'Delete</button>' +
-      '</div>' +
-      '</div>' +
-      '</div>' +
-      '</div>' +
-      '</div>';
-  return formattedHTML;
+
+  // Fill in template fields with correct information.
+  const templateScope = document.querySelector('#receipt-card-template');
```
bryyeh, line 70: When is it better to use querySelector() vs getElementById()? I'm just curious since I'm not as familiar with javascript.

src/main/java/servlets/SearchServlet.java
emilyvera: Thanks! I've never seen that notation before but that does make it cleaner

src/main/webapp/js/index.js
emilyvera: No, receipts-display is the container that allows the inner receipt cards to be responsive :/

src/main/webapp/js/index.js
emilyvera: I'm not sure that there really is a difference (stack overflow says querySelector is newer and the browser support might be different). I was following a tutorial to use templates and it used this method, so I just went with it

src/main/webapp/js/index.js
```diff
@@ -13,36 +13,29 @@
...
-  const receiptsDisplay = document.getElementById('receipts-display');
-  const numChildNodes = receiptsDisplay.childNodes.length;
-
-  // Make sure not to delete first 3 childNodes
-  for (let i = numChildNodes; i > 3; i--) {
-    receiptsDisplay.removeChild(receiptsDisplay.lastChild);
-  }
+function clearExistingDisplay() {
+  const elements = document.querySelectorAll(
+      '.col-md-6, .col-md-12.text-center.error-message');
```
bryyeh, line 31: I would add a comment or have this string be in a descriptive variable. Otherwise it's not clear why you are querying for this.

src/main/webapp/js/index.js
acheng1: I think this is moot now that you are using the querySelector, but the template itself I think can be placed anywhere in the document, but once you do clone the template, it does have to be appended as a child of receipts-display

src/main/webapp/js/index.js
```diff
@@ -12,8 +12,92 @@
...
+  clearExistingDisplay();
+  displayReceipts(label, receipts);
+}
+
+/**
+ * Clears out receipts display including old receipts and error messages.
+ */
+function clearExistingDisplay() {
+  const elements = document.querySelectorAll(
+      '.col-md-6, .col-md-12.text-center.error-message');
```
acheng1, line 31: This method is a lot cleaner now, but referencing CSS classes can make this brittle (e.g., if you decide to change the style, someone needs to make sure to update this reference in the javascript. I do think you could declare the templates outside of the `receipts-display` div element and then you would be able to just do `document.getElementById('receipts-display').innerHtml = '';` to reset things.

src/main/webapp/js/index.js
```diff
@@ -12,8 +12,92 @@
...
+      document.querySelector('#receipt-card-template').content.cloneNode(true);
+
+  // Fill in template fields with correct information.
+  receiptCardClone.querySelector('#timestamp').innerText = receipt.timestamp;
+  receiptCardClone.querySelector('#store-name').innerText = receipt.store;
+  receiptCardClone.querySelector('#total').innerText =
+      'Total: $' + receipt.price;
+  receiptCardClone.querySelector('#c1').innerText = categories[0];
+  receiptCardClone.querySelector('#c2').innerText = categories[1];
+  receiptCardClone.querySelector('#c3').innerText = categories[2];
```
acheng1, line 73: Presumably there may be less than 3 categories, you might want to loop through the categories using  `receiptCardClone.querySelector('#c' + i).innerText = ...` up to the 3 categories.

src/main/webapp/js/index.js
emilyvera: Oh I see what you mean now. I didn't know templates could be placed anywhere in the html and then attached where you want them. That would've saved me a lot of headache

PR #12, opened 2020-07-06
Integrate upload servlet with shared receipt analysis class and add error handling.

basaam0:


src/main/java/servlets/UploadReceiptServlet.java
```diff
@@ -173,4 +164,65 @@ private static boolean isValidFilename(String filename) {
...
+  /**
+   * Get the base URL of the application for either the dev server or deployment.
+   */
+  private String getBaseUrl(HttpServletRequest request) {
+    String baseUrl = request.getScheme() + "://" + request.getServerName() + ":"
+        + request.getServerPort() + request.getContextPath();
+
+    // URL for dev server.
+    if (baseUrl.equals("http://0.0.0.0:80")) {
+      return "https://8080-9333ac09-1dd5-4f7f-8a7a-34f16c364c6b.us-east1.cloudshell.dev";
```
avagarcia0, line 205: Is it possible to avoid hardcoding this URL? It'll be different for Emily and me.

src/main/java/servlets/UploadReceiptServlet.java
```diff
@@ -173,4 +164,65 @@ private static boolean isValidFilename(String filename) {
...
+    }
+
+    return baseUrl;
+  }
+
+  public class InvalidFileException extends Exception {
```
avagarcia0, line 211: Should we place these exceptions somewhere else? I might want to make use of some of them in ReceiptAnalysis, and it wouldn't make sense to import them from a servlet that depends on ReceiptAnalysis.

src/main/java/servlets/UploadReceiptServlet.java
basaam0: For this commit, accessing receipt images on the back-end from the URL was not working on the dev server due to authentication issues with Cloud Shell. 

I looked into a few alternative solutions like getting resources through the servlet context and using a request dispatcher, and considering that user authentication checks will be added to the servlets for the MVP, I think that the best option is fetching the bytes with the blob key. The latest commit uses the imageUrl in deployment and the blob key for the dev server to access the receipt image from the shared ReceiptAnalysis class.

src/main/java/servlets/UploadReceiptServlet.java
basaam0: I have moved the ReceiptAnalysisException into the shared class. I think the other two exceptions are specific to this class.

src/main/java/servlets/ReceiptAnalysis.java
```diff
@@ -53,6 +66,29 @@ private static ByteString readImageBytes(String url) throws IOException {
...
+    long blobSize = blobInfo.getSize();
+
+    ByteArrayOutputStream outputBytes = new ByteArrayOutputStream();
+    int fetchSize = BlobstoreService.MAX_BLOB_FETCH_SIZE;
+    long currentByteIndex = 0;
+
+    // Fetch all the bytes from the blob in fragments of the maximum fetch size.
+    while (currentByteIndex < blobSize) {
+      // End index is inclusive, so subtract 1 to get fetchSize bytes.
+      byte[] b =
```
emilyvera, line 82: maybe name this bytes instead of just b

src/main/java/servlets/UploadReceiptServlet.java
```diff
@@ -75,69 +78,58 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
-      response.sendError(HttpServletResponse.SC_BAD_REQUEST, "No JPEG file uploaded.");
+    Entity receipt = null;
+
+    try {
+      receipt = createReceiptEntity(request);
+    } catch (FileNotSelectedException | InvalidFileException e) {
+      logger.warning(e.toString());
+      response.setStatus(HttpServletResponse.SC_BAD_REQUEST);
+      response.getWriter().println(e.toString());
+      return;
```
emilyvera, line 89: I thought we were gonna show a message to the user here too, or was that somewhere else?

basaam0: The code for showing the message is in upload.js (L41)

src/main/java/servlets/ReceiptAnalysis.java
```diff
@@ -53,6 +66,29 @@ private static ByteString readImageBytes(String url) throws IOException {
...
+
+    ByteArrayOutputStream outputBytes = new ByteArrayOutputStream();
+    int fetchSize = BlobstoreService.MAX_BLOB_FETCH_SIZE;
+    long currentByteIndex = 0;
+
+    // Fetch all the bytes from the blob in fragments of the maximum fetch size.
+    while (currentByteIndex < blobSize) {
+      // End index is inclusive, so subtract 1 to get fetchSize bytes.
+      byte[] bytes =
+          blobstoreService.fetchData(blobKey, currentByteIndex, currentByteIndex + fetchSize - 1);
```
bryyeh, line 83: I assume it's okay for the end index to be greater than the blobSize?

src/main/java/servlets/UploadReceiptServlet.java
```diff
@@ -167,10 +158,72 @@ private static boolean isValidFilename(String filename) {
...
 
+  /**
+   * Extracts the raw text from the image with the Cloud Vision API. Returns a receipt
+   * entity populated with the extracted fields.
+   */
+  private Entity analyzeReceiptImage(BlobKey blobKey, HttpServletRequest request)
+      throws ReceiptAnalysisException {
+    String imageUrl = getBlobServingUrl(blobKey);
+    String baseUrl = getBaseUrl(request);
+    String DEV_SERVER_BASE_URL = "http://0.0.0.0:80";
```
bryyeh, line 169: This should be a class member variable that is private static final.

src/main/java/servlets/ReceiptAnalysis.java
basaam0: Yes, fetchData() will not read past the blob with the specified blob key.

PR #13, opened 2020-07-06
Add style and media queries to upload page.

basaam0:
style1
style2

### DIFF
--- a/src/main/webapp/css/upload.css
+++ b/src/main/webapp/css/upload.css
@@ -34,9 +34,20 @@ input {
   width: 7em;
 }
 
+.custom-file {
+  font-size: 0.5rem;
+}
+
+#receipt-image-label {
+  margin: 0;
+}
+
 #receipt-image-input {
-  display: block;
-  padding: 0;
+  display: none;
+}
+
+.custom-file-label {
+  font-size: 1rem;
 }
 
 small {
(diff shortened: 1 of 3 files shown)


src/main/webapp/css/upload.css
```diff
@@ -1,9 +1,99 @@
...
+#upload-form,
+input {
+  font-size: 40px;
+}
+
+#upload-form {
+  background: linear-gradient(#e3ffe7 0%, #d9e7ff 100%);
+  border: solid 2px transparent;
+  border-radius: 30px;
+  padding: 4px;
```
emilyvera, line 24: this is minor but these should be alphabetized (switch margin and padding)

src/main/webapp/js/upload.js
```diff
@@ -55,16 +61,77 @@ async function fetchBlobstoreUrl() {
...
+    const fileName = fileInput.value.split('\\').pop();
+    fileLabel.innerText = fileName;
+  } else {
+    fileLabel.innerText = DEFAULT_FILE_LABEL;
+  }
+}
+
 /**
  * Displays an error message if the user selects a file larger than 5 MB.
+ * Returns true if the user selected a file with a smaller size.
```
emilyvera, line 119: It would be nice to style the "returns true" line as described in the JS style guide:

@return {boolean} Whether the file size is less than 5 MB.

https://google.github.io/styleguide/jsguide.html#jsdoc-method-and-function-comments

src/main/webapp/js/upload.js
```diff
@@ -48,23 +54,85 @@ async function uploadReceipt(event) {
...
+          style: 'currency',
+          currencyDisplay: 'symbol',
+        });
+  } else {
+    event.target.value = '';
+  }
+}
+
+/**
+ * Add the selected file name to the input label and check that the size of the
```
acheng1, line 102: nitnit: Adds (thirdperson)

PR #15, opened 2020-07-08
Add unit tests for receipt upload servlet.

basaam0:


src/test/java/UploadReceiptServletTest.java
```diff
@@ -0,0 +1,331 @@
...
+  }
+
+  @Test
+  public void doGetReturnsBlobstoreUploadUrl() throws IOException {
+    StringWriter stringWriter = new StringWriter();
+    PrintWriter writer = new PrintWriter(stringWriter);
+    when(response.getWriter()).thenReturn(writer);
+
+    UploadOptions uploadOptions =
+        UploadOptions.Builder.withMaxUploadSizeBytesPerBlob(MAX_UPLOAD_SIZE_BYTES);
```
acheng1, line 143: nitnit: is there an extra newline here?

src/test/java/UploadReceiptServletTest.java
```diff
@@ -0,0 +1,331 @@
...
+  }
+
+  @After
+  public void tearDown() {
+    helper.tearDown();
+  }
+
+  @Test
+  public void doGetReturnsBlobstoreUploadUrl() throws IOException {
+    StringWriter stringWriter = new StringWriter();
```
acheng1, line 138: We typically structure the tests into 3 distinct blocks: arrange, act, assert: every test must arrange the state of the world to test, act on the class under test by calling methods, and assert that the world is in the expected state afterward.

```
<Arrange statements>
when(... 
...

<Act statements>
servlet.doGet(...
...

<Assert statements>
Assert....
...
<newline>
```

src/test/java/UploadReceiptServletTest.java
```diff
@@ -0,0 +1,331 @@
...
+    PrintWriter writer = new PrintWriter(stringWriter);
+    when(response.getWriter()).thenReturn(writer);
+
+    UploadOptions uploadOptions =
+        UploadOptions.Builder.withMaxUploadSizeBytesPerBlob(MAX_UPLOAD_SIZE_BYTES);
+    when(blobstoreService.createUploadUrl("/upload-receipt", uploadOptions)).thenReturn(UPLOAD_URL);
+
+    servlet.doGet(request, response);
+    writer.flush();
+    Assert.assertTrue(stringWriter.toString().contains(UPLOAD_URL));
```
acheng1, line 148: Based on the structure mentioned above, I'd add a newline before the Assert.

src/test/java/UploadReceiptServletTest.java
```diff
@@ -0,0 +1,331 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.sps;
+
+import static org.mockito.Mockito.*;
```
avagarcia0, line 17: From the Google Java style guide:

>  Wildcard imports, static or otherwise, are not used.

https://google.github.io/styleguide/javaguide.html#s3.3.1-wildcard-imports

src/test/java/UploadReceiptServletTest.java
```diff
@@ -0,0 +1,331 @@
...
+    when(blobInfoFactory.loadBlobInfo(BLOB_KEY)).thenReturn(blobInfo);
+
+    when(request.getParameter("label")).thenReturn(LABEL);
+
+    // Stub request with URL components.
+    when(request.getScheme()).thenReturn(LIVE_SERVER_SCHEME);
+    when(request.getServerName()).thenReturn(LIVE_SERVER_NAME);
+    when(request.getServerPort()).thenReturn(LIVE_SERVER_PORT);
+    when(request.getContextPath()).thenReturn(LIVE_SERVER_CONTEXT_PATH);
+    String absoluteUrl = LIVE_SERVER_BASE_URL + IMAGE_URL;
```
avagarcia0, line 168: I think it would be better to have an ABSOLUTE_URL constant constructed from the other two constants rather than constructing that string in two different places.

src/test/java/UploadReceiptServletTest.java
```diff
@@ -0,0 +1,331 @@
...
+import com.google.appengine.api.datastore.PreparedQuery;
+import com.google.appengine.api.datastore.Query;
+import com.google.appengine.api.datastore.Text;
+import com.google.appengine.tools.development.testing.LocalDatastoreServiceTestConfig;
+import com.google.appengine.tools.development.testing.LocalServiceTestHelper;
+import com.google.gson.Gson;
+import com.google.sps.data.AnalysisResults;
+import com.google.sps.servlets.ReceiptAnalysis;
+import com.google.sps.servlets.UploadReceiptServlet;
+import java.io.*;
```
avagarcia0, line 40: This wildcard needs removed as well.

avagarcia0, line 40: This wildcard needs removed as well.

src/test/java/UploadReceiptServletTest.java
```diff
@@ -0,0 +1,331 @@
...
+import java.time.ZoneId;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Calendar;
+import java.util.Date;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.logging.Logger;
+import javax.servlet.http.*;
```
avagarcia0, line 52: This wildcard needs removed as well.

src/main/java/servlets/UploadReceiptServlet.java
```diff
@@ -53,8 +54,33 @@
...
+
+  @Override
+  public void init() {
+    // Logs to System.err by default.
+    logger = Logger.getLogger(UploadReceiptServlet.class.getName());
+    blobstoreService = BlobstoreServiceFactory.getBlobstoreService();
+    blobInfoFactory = new BlobInfoFactory();
+    datastore = DatastoreServiceFactory.getDatastoreService();
+    clock = Clock.systemDefaultZone();
+  }
```
acheng1, line 84: nit: since this is a refactoring, for future reference, I'd put this in a separate pull request.

src/test/java/UploadReceiptServletTest.java
```diff
@@ -0,0 +1,335 @@
...
+    PrintWriter writer = new PrintWriter(stringWriter);
+    when(response.getWriter()).thenReturn(writer);
+
+    UploadOptions uploadOptions =
+        UploadOptions.Builder.withMaxUploadSizeBytesPerBlob(MAX_UPLOAD_SIZE_BYTES);
+    when(blobstoreService.createUploadUrl("/upload-receipt", uploadOptions)).thenReturn(UPLOAD_URL);
+
+    servlet.doGet(request, response);
+    writer.flush();
+    Assert.assertTrue(stringWriter.toString().contains(UPLOAD_URL));
```
acheng1, line 154: We try to structure the test into 3 distinct blocks with a newline between them: arrange, act, assert: every test must arrange the state of the world to test, act on the class under test by calling methods, and assert that the world is in the expected state afterward. So, I would try to group the statements into these 3 blocks.

acheng1, line 154: We try to structure the test into 3 distinct blocks with a newline between them: arrange, act, assert: every test must arrange the state of the world to test, act on the class under test by calling methods, and assert that the world is in the expected state afterward. So, I would try to group the statements into these 3 blocks and add a newline between each (newline between `writer.flush();` and `Assert.assertTrue...;`

```
<arrange statements>
...
<newline>
<act statements>
...
<newline>
<assert statements>
...
```

src/test/java/UploadReceiptServletTest.java
```diff
@@ -0,0 +1,335 @@
...
+
+    Assert.assertTrue(stringWriter.toString().contains(INVALID_FILE_WARNING));
+    verify(logger).warning(INVALID_FILE_WARNING);
+    verify(response).setStatus(HttpServletResponse.SC_BAD_REQUEST);
+
+    verify(blobstoreService).delete(BLOB_KEY);
+  }
+
+  @Test
+  public void doPostThrowsIfReceiptAnalysisFails() throws IOException {
```
acheng1, line 301: Great test coverage!

src/test/java/UploadReceiptServletTest.java
basaam0: Should these be the only newlines or can there be more within each block?

acheng1: There can be more within each block.

src/main/java/servlets/UploadReceiptServlet.java
```diff
@@ -181,6 +204,7 @@ private Entity analyzeReceiptImage(BlobKey blobKey, HttpServletRequest request)
         results = ReceiptAnalysis.serveImageText(absoluteUrl);
       }
     } catch (IOException e) {
+      blobstoreService.delete(blobKey);
```
acheng1, line 207: Also this too, for future reference, I would include in a different pull request.

src/test/java/UploadReceiptServletTest.java
```diff
@@ -0,0 +1,335 @@
...
+  private final LocalServiceTestHelper helper =
+      new LocalServiceTestHelper(new LocalDatastoreServiceTestConfig());
+
+  @Mock private Logger logger;
+  @Mock private BlobstoreService blobstoreService;
+  @Mock private BlobInfoFactory blobInfoFactory;
+  @Mock private HttpServletRequest request;
+  @Mock private HttpServletResponse response;
+
+  private static final String FILE_NOT_SELECTED_LIVE_SERVER_WARNING =
```
acheng1, line 85: Generally we would try to order the file as:

We recommend a high-level ordering like this (notice that visible static fields and static methods are grouped together before instance fields):

Visible static fields (usually final)
Visible static methods
"Regular" instance fields (comprising object state, not derived)
Constructors (whether visible or not)
Visible instance methods

src/main/java/servlets/UploadReceiptServlet.java
```diff
@@ -53,8 +54,33 @@
...
   // Matches JPEG image filenames.
   private static final Pattern validFilename = Pattern.compile("([^\\s]+(\\.(?i)(jpe?g))$)");
-  // Logs to System.err by default.
-  private static final Logger logger = Logger.getLogger(UploadReceiptServlet.class.getName());
+
+  private static Logger logger;
+  private static BlobstoreService blobstoreService;
+  private static BlobInfoFactory blobInfoFactory;
+  private static DatastoreService datastore;
+  private static Clock clock;
```
acheng1, line 62: I would make these final but not static. I would also be careful about making instance variables static (see http://www.object-oriented-security.org/lets-argue/singletons).

src/main/java/servlets/UploadReceiptServlet.java
```diff
@@ -53,8 +54,33 @@
   private static final String DEV_SERVER_BASE_URL = "http://0.0.0.0:80";
   // Matches JPEG image filenames.
   private static final Pattern validFilename = Pattern.compile("([^\\s]+(\\.(?i)(jpe?g))$)");
-  // Logs to System.err by default.
-  private static final Logger logger = Logger.getLogger(UploadReceiptServlet.class.getName());
+
+  private static Logger logger;
```
acheng1, line 58: Okay to keep one instance of logger (static) but I would also make this final.

basaam0: Marking the logger both final and static will prevent assigning it a value in the constructor since it is called for every instance. Should I only mark it as final then?

acheng1: I don't think it's necessary to create a new instance of the logger for each request so I would keep it static and final. I also think it's not super necessary to check the log message since you check the error status already. If you do want to check the log message though, there are options  to testing the logging by creating a mock appender.

PR #18, opened 2020-07-09
Delete receipt servlet

emilyvera:
- Delete button for each receipt card deletes receipt entity from DOM and database.
- Alert will pop-up to confirm deletion.
- Added tests and a test utility class.

### DIFF
--- a/makefile
+++ b/makefile
@@ -12,6 +12,7 @@ pretty: node_modules
 	$(PRETTIER) --write src/main/webapp/css/*.css
 	find src/main/java -iname *.java | xargs $(CLANG_FORMAT) -i
 	find src/main/webapp -iname *.js | xargs $(CLANG_FORMAT) -i
+	find src/test/java -iname *.java | xargs $(CLANG_FORMAT) -i
 
 validate: node_modules
 	$(HTML_VALIDATE) src/main/webapp/*.html
--- a/makefile
+++ b/makefile
@@ -10,9 +10,8 @@ node_modules:
 pretty: node_modules
 	$(PRETTIER) --write src/main/webapp/*.html
 	$(PRETTIER) --write src/main/webapp/css/*.css
-	find src/main/java -iname *.java | xargs $(CLANG_FORMAT) -i
+	find src/main/java src/test/java -iname *.java | xargs $(CLANG_FORMAT) -i
 	find src/main/webapp -iname *.js | xargs $(CLANG_FORMAT) -i
-	find src/test/java -iname *.java | xargs $(CLANG_FORMAT) -i
 
 validate: node_modules
 	$(HTML_VALIDATE) src/main/webapp/*.html
--- a/src/main/java/servlets/DeleteReceiptServlet.java
+++ b/src/main/java/servlets/DeleteReceiptServlet.java
@@ -27,7 +27,6 @@ import javax.servlet.http.HttpServletResponse;
 /** Servlet responsible for deleting a single receipt. */
 @WebServlet("/delete-receipt")
 public class DeleteReceiptServlet extends HttpServlet {
-
   private static DatastoreService datastore;
 
   public DeleteReceiptServlet() {}
@@ -40,7 +39,7 @@ public class DeleteReceiptServlet extends HttpServlet {
   public void init() {
     datastore = DatastoreServiceFactory.getDatastoreService();
   }
-    
+
   @Override
   public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
     long id = Long.parseLong(request.getParameter("id"));
--- a/src/main/java/servlets/DeleteReceiptServlet.java
+++ b/src/main/java/servlets/DeleteReceiptServlet.java
@@ -44,7 +44,7 @@ public class DeleteReceiptServlet extends HttpServlet {
   public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
     long id = Long.parseLong(request.getParameter("id"));
 
-    Key receiptEntityKey = KeyFactory.createKey("Receipt", id);
-    datastore.delete(receiptEntityKey);
+    Key key = KeyFactory.createKey("Receipt", id);
+    datastore.delete(key);
   }
 }
--- a/src/main/java/servlets/DeleteReceiptServlet.java
+++ b/src/main/java/servlets/DeleteReceiptServlet.java
@@ -50,17 +50,12 @@ public class DeleteReceiptServlet extends HttpServlet {
       id = Long.parseLong(request.getParameter("id"));
     } catch (NumberFormatException exception) {
       response.setStatus(HttpServletResponse.SC_BAD_REQUEST);
-      response.getWriter().println("Invalid ID: Receipt unable to be deleted at this time, please try again.");
+      response.getWriter().println(
+          "Invalid ID: Receipt unable to be deleted at this time, please try again.");
       return;
     }
 
     Key key = KeyFactory.createKey("Receipt", id);
-
-    try {
-      datastore.delete(key);
-    } catch (DatastoreFailureException exception) {
-      response.setStatus(HttpServletResponse.SC_INTERNAL_SERVER_ERROR);
-      response.getWriter().println("Internal Datastore Error: Receipt unable to be deleted at this time, please try again.");
-    }
+    datastore.delete(key);
   }
 }
--- a/src/main/java/servlets/DeleteReceiptServlet.java
+++ b/src/main/java/servlets/DeleteReceiptServlet.java
@@ -29,17 +29,14 @@ import javax.servlet.http.HttpServletResponse;
 /** Servlet responsible for deleting a single receipt. */
 @WebServlet("/delete-receipt")
 public class DeleteReceiptServlet extends HttpServlet {
-  private static DatastoreService datastore;
+  private final DatastoreService datastore;
 
-  public DeleteReceiptServlet() {}
+  public DeleteReceiptServlet() {
+    datastore = DatastoreServiceFactory.getDatastoreService();
+  }
 
   public DeleteReceiptServlet(DatastoreService datastore) {
     this.datastore = datastore;
-  }
-
-  @Override
-  public void init() {
-    datastore = DatastoreServiceFactory.getDatastoreService();
   }
 
   @Override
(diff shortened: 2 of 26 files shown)


src/main/java/servlets/DeleteReceiptServlet.java
```diff
@@ -0,0 +1,50 @@
...
+  @Override
+  public void init() {
+    datastore = DatastoreServiceFactory.getDatastoreService();
+  }
+
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    long id = Long.parseLong(request.getParameter("id"));
+
+    Key receiptEntityKey = KeyFactory.createKey("Receipt", id);
```
basaam0, line 47: A best practice for naming variables from tott/566 is to omit words that are obvious given a variable's type declaration. 

https://drive.google.com/file/d/1-VB6otFQRR2SOESwAXRCHiHiZ4tx11iw/view

src/main/webapp/js/index.js
```diff
@@ -44,11 +44,29 @@ function displayReceipts(label, receipts) {
...
+ */
+function createErrorMessageElement(label) {
+  // Clone error message from template.
+  const errorMessageClone =
+      document.querySelector('#error-message-template').content.cloneNode(true);
+
+  // Fill in template fields with correct information.
+  errorMessageClone.querySelector('h3').innerText =
+      'Sorry, no results found for "' + label +
+      '". Please try your search again or try a different query.';
```
basaam0, line 59: Embedding the label in the string using template literals would make this easier to read.

src/main/webapp/js/index.js
```diff
@@ -70,24 +88,38 @@ function createReceiptCardElement(receipt) {
...
-
-  // Fill in template fields with correct information.
-  errorMessageClone.querySelector('h3').innerText =
-      'Sorry, no results found for "' + label +
-      '". Please try your search again or try a different query.';
+function attachDeleteButtonEventListener(receipt) {
+  // Get all delete buttons; most recently attached will be at the end.
+  const allDeleteButtons =
+      document.querySelector('#receipts-display').querySelectorAll('#delete');
+  const mostRecentDeleteButton = allDeleteButtons[allDeleteButtons.length - 1];
```
basaam0, line 104: Maybe including `receiptCardClone` as an argument and using a query selector on it would be more efficient than getting all the delete buttons.

src/test/java/DeleteReceiptServletTest.java
```diff
@@ -0,0 +1,86 @@
...
+  @Mock private DatastoreService datastore;
+  @Mock private DeleteReceiptServlet servlet;
+  @Mock private HttpServletRequest request;
+  @Mock private HttpServletResponse response;
+
+  @Before
+  public void setUp() {
+    MockitoAnnotations.initMocks(this);
+    helper.setUp();
+    datastore = DatastoreServiceFactory.getDatastoreService();
```
basaam0, line 58: Datastore does not need the `@Mock` annotation since a local version is being used.

src/test/java/TestUtils.java
```diff
@@ -0,0 +1,51 @@
...
+    receiptEntity.setProperty("imageUrl", "img/walmart-receipt.jpg");
+    receiptEntity.setProperty("price", 26.12);
+    receiptEntity.setProperty("store", "Walmart");
+    receiptEntity.setProperty("label", "test");
+    receiptEntity.setProperty(
+        "categories", new HashSet<>(Arrays.asList("Cappuccino", "Sandwich", "Lunch")));
+    receiptEntity.setProperty("rawText", "");
+
+    // Add receipt to datastore and return ID.
+    Key receiptKey = datastore.put(receiptEntity);
```
basaam0, line 48: > A best practice for naming variables from tott/566 is to omit words that are obvious given a variable's type declaration.
> 
> https://drive.google.com/file/d/1-VB6otFQRR2SOESwAXRCHiHiZ4tx11iw/view

src/test/java/TestUtils.java
```diff
@@ -0,0 +1,51 @@
...
+    receiptEntity.setProperty("userId", 1);
+    receiptEntity.setProperty("timestamp", 6292020);
+    receiptEntity.setProperty("blobkey", new BlobKey("Test"));
+    receiptEntity.setProperty("imageUrl", "img/walmart-receipt.jpg");
+    receiptEntity.setProperty("price", 26.12);
+    receiptEntity.setProperty("store", "Walmart");
+    receiptEntity.setProperty("label", "test");
+    receiptEntity.setProperty(
+        "categories", new HashSet<>(Arrays.asList("Cappuccino", "Sandwich", "Lunch")));
+    receiptEntity.setProperty("rawText", "");
```
basaam0, line 45: Not sure if this applies here but I would move the hardcoded properties to class constants in case they are used again.

src/main/webapp/js/index.js
emilyvera: receiptCardClone is a DocumentFragment, so after it's attached to the DOM, the DocumentFragment is left empty. Trying to access the delete button from receiptCardClone would result in a null exception on the click event

https://developer.mozilla.org/en-US/docs/Web/API/DocumentFragment#usage_notes

makefile
```diff
@@ -12,6 +12,7 @@ pretty: node_modules
 	$(PRETTIER) --write src/main/webapp/css/*.css
 	find src/main/java -iname *.java | xargs $(CLANG_FORMAT) -i
 	find src/main/webapp -iname *.js | xargs $(CLANG_FORMAT) -i
+	find src/test/java -iname *.java | xargs $(CLANG_FORMAT) -i
```
basaam0, line 15: With `find`, it's possible to specify multiple directories to avoid repeating the same command.

`find src/main/java src/test/java -iname *.java | xargs $(CLANG_FORMAT) -i`

avagarcia0:
Just so you know I'm waiting for changes to be pushed before I review this (I assume that the resolved comments mean you've made the changes already and just haven't pushed them yet)

src/main/webapp/js/index.js
bryyeh: Could the receipt card element contain the receipt id? Then that can be used to locate the receipt to remove from the DOM.

src/test/java/TestUtils.java
```diff
@@ -0,0 +1,61 @@
...
+
+import com.google.appengine.api.blobstore.BlobKey;
+import com.google.appengine.api.datastore.DatastoreService;
+import com.google.appengine.api.datastore.Entity;
+import com.google.appengine.api.datastore.Key;
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.HashSet;
+
+/** Class that contains helpful methods used for testing. */
```
bryyeh, line 25: I love you made a TestUtils class!

src/test/java/DeleteReceiptServletTest.java
```diff
@@ -0,0 +1,127 @@
...
+    PreparedQuery results = datastore.prepare(query);
+    Assert.assertEquals(1, results.countEntities(FetchOptions.Builder.withDefaults()));
+
+    // Perform doPost - this should delete the receipt.
+    when(request.getParameter("id")).thenReturn(String.valueOf(id));
+    servlet.doPost(request, response);
+
+    // Make sure receipt is deleted by checking if there are no entities returned.
+    Key desiredKey = KeyFactory.createKey("Receipt", id);
+    Filter matchingKeys = new FilterPredicate("__key__", FilterOperator.EQUAL, desiredKey);
```
bryyeh, line 96: I think it's simpler to check if the whole datastore is empty since there was only one receipt inserted.

src/main/java/servlets/DeleteReceiptServlet.java
```diff
@@ -0,0 +1,61 @@
...
+      id = Long.parseLong(request.getParameter("id"));
+    } catch (NumberFormatException exception) {
+      response.setStatus(HttpServletResponse.SC_BAD_REQUEST);
+      response.getWriter().println(
+          "Invalid ID: Receipt unable to be deleted at this time, please try again.");
+      return;
+    }
+
+    Key key = KeyFactory.createKey("Receipt", id);
+    datastore.delete(key);
```
bryyeh, line 59: If the datastore delete fails, it would be good to have some handling to tell the user.

src/main/java/servlets/DeleteReceiptServlet.java
```diff
@@ -0,0 +1,61 @@
...
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+/** Servlet responsible for deleting a single receipt. */
+@WebServlet("/delete-receipt")
+public class DeleteReceiptServlet extends HttpServlet {
+  private static DatastoreService datastore;
+
+  public DeleteReceiptServlet() {}
```
acheng1, line 34: Is this constructor actually needed?

src/main/webapp/js/index.js
acheng1: The `appendChild()` call above should return the node that was appended so you could pass that in into` attachDeleteButtonEventListener()` and then query the delete button within that node. Something like:
```
  // Attach receipt card clone to parent div.
  const node = document.getElementById('receipts-display').appendChild(receiptCardClone);

  // Attach listener to trigger the deletion of this receipt.
  attachDeleteButtonEventListener(node, receipt);
```

emilyvera: @acheng1 I tried the appendChild() method but according to the documentation: 

> The returned value is the appended child (aChild), except when aChild is a DocumentFragment, in which case the empty DocumentFragment is returned.

So the issue of the node being null on the button click would still exist. I found Bryan's suggestion worked well though!

src/main/webapp/js/index.js
```diff
@@ -67,27 +85,37 @@ function createReceiptCardElement(receipt) {
   }
 
   receiptCardClone.querySelector('img').src = receipt.imageUrl;
+  receiptCardClone.querySelector('.col-md-6').id = receipt.id;
```
acheng1, line 88: nit: can we use id's for these, just seems a little brittle to rely on css as I can see the css as  being something that could change later.

emilyvera: Could you clarify what you mean by this? Like change col-md-6?

PR #21, opened 2020-07-09
Create login status servlet.

basaam0:
The login status servlet checks whether the user is logged in and retrieves information such as a login URL, logout URL, and email address.

### DIFF
--- a/src/main/java/data/Account.java
+++ b/src/main/java/data/Account.java
@@ -28,7 +28,6 @@ public class Account {
   public Account(String loginUrl) {
     this.loggedIn = false;
     this.loginUrl = loginUrl;
-
     this.logoutUrl = this.email = null;
   }
 
@@ -39,7 +38,6 @@ public class Account {
     this.loggedIn = true;
     this.logoutUrl = logoutUrl;
     this.email = email;
-
     this.loginUrl = null;
   }
 }
--- a/src/main/java/data/Account.java
+++ b/src/main/java/data/Account.java
@@ -15,7 +15,7 @@
 package com.google.sps.data;
 
 /** Information about the logged-in user. */
-public class AuthenticationInformation {
+public class Account {
   // Note: types cannot be Optional due to JSON conversion.
   private final boolean loggedIn;
   private final String loginUrl;
@@ -25,7 +25,7 @@ public class AuthenticationInformation {
   /**
    * Creates a logged out user with a URL to log in.
    */
-  public AuthenticationInformation(String loginUrl) {
+  public Account(String loginUrl) {
     this.loggedIn = false;
     this.loginUrl = loginUrl;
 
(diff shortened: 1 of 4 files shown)


src/test/java/LoginStatusServletTest.java
```diff
@@ -0,0 +1,125 @@
...
+    StringWriter stringWriter = new StringWriter();
+    PrintWriter writer = new PrintWriter(stringWriter);
+    when(response.getWriter()).thenReturn(writer);
+
+    servlet.doGet(request, response);
+    writer.flush();
+
+    String actual = getJsonResponse(stringWriter);
+    String expected = new Gson().toJson(LOGGED_IN_INFORMATION);
+    Assert.assertEquals(expected, actual);
```
avagarcia0, line 116: Since you are constructing the expected string and testing the actual response, I think it would be best to only manipulate the expected string. In this case, you could just add a newline character to the end of it.

src/main/java/servlets/LoginStatusServlet.java
```diff
@@ -0,0 +1,75 @@
...
+      String logoutUrl = userService.createLogoutURL(LOGIN_PAGE_URL);
+      String email = userService.getCurrentUser().getEmail();
+      authentication = new AuthenticationInformation(logoutUrl, email);
+    } else {
+      String loginUrl = userService.createLoginURL(HOME_PAGE_URL);
+      authentication = new AuthenticationInformation(loginUrl);
+    }
+
+    // Convert the authentication information to JSON.
+    String json = convertToJson(authentication);
```
bryyeh, line 60: Would new Gson().toJson(authentication) take out the need for the helper method?

src/main/java/data/AuthenticationInformation.java
```diff
@@ -0,0 +1,40 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.sps.data;
+
+/** Information about the logged-in user. */
+public class AuthenticationInformation {
```
bryyeh, line 18: nitnit: Is there a shorter name we could think of?

src/main/java/data/Account.java
```diff
@@ -15,26 +15,31 @@
...
+  private final String email;
 
   /**
    * Creates a logged out user with a URL to log in.
    */
-  public AuthenticationInformation(String loginUrl) {
+  public Account(String loginUrl) {
+    this.loggedIn = false;
     this.loginUrl = loginUrl;
+
```
bryyeh, line 31: Don't think the newline is necessary. Same on L42.

PR #22, opened 2020-07-10
Insert user ID field in the receipt entity and update upload servlet unit tests.

basaam0:
Uses a local UserService for unit testing

### DIFF
--- a/src/main/java/servlets/UploadReceiptServlet.java
+++ b/src/main/java/servlets/UploadReceiptServlet.java
@@ -112,6 +112,11 @@ public class UploadReceiptServlet extends HttpServlet {
       response.setStatus(HttpServletResponse.SC_BAD_REQUEST);
       response.getWriter().println(e.toString());
       return;
+    } catch (UserNotLoggedInException e) {
+      logger.warning(e.toString());
+      response.setStatus(HttpServletResponse.SC_FORBIDDEN);
+      response.getWriter().println(e.toString());
+      return;
     } catch (ReceiptAnalysisException e) {
       logger.warning(e.toString());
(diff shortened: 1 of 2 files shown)


src/main/java/servlets/UploadReceiptServlet.java
```diff
@@ -60,20 +62,23 @@
...
     this.blobstoreService = BlobstoreServiceFactory.getBlobstoreService();
     this.blobInfoFactory = new BlobInfoFactory();
     this.datastore = DatastoreServiceFactory.getDatastoreService();
+    this.userService = UserServiceFactory.getUserService();
     this.clock = Clock.systemDefaultZone();
   }
 
   public UploadReceiptServlet(BlobstoreService blobstoreService, BlobInfoFactory blobInfoFactory,
-      DatastoreService datastore, Clock clock) {
+      DatastoreService datastore, UserService userService, Clock clock) {
```
bryyeh, line 77: Does it make a difference if the UserService is initialized in the constructor vs passed in as a parameter in the test?

src/main/java/servlets/UploadReceiptServlet.java
```diff
@@ -129,12 +134,14 @@ private Entity createReceiptEntity(HttpServletRequest request)
     BlobKey blobKey = getUploadedBlobKey(request, "receipt-image");
     long timestamp = clock.instant().toEpochMilli();
     String label = request.getParameter("label");
+    String userId = userService.getCurrentUser().getUserId();
```
bryyeh, line 137: Should we have a check to ensure the user is logged in?

PR #23, opened 2020-07-10
Replace hard-coded store name and total price fields in receipt entity.

basaam0:
- Send store name and total price from the upload form to the upload servlet
- Insert fields in Datastore entity
- Update upload servlet unit tests
- Add unit test for rounding total price

### DIFF
--- a/src/main/java/servlets/UploadReceiptServlet.java
+++ b/src/main/java/servlets/UploadReceiptServlet.java
@@ -129,12 +129,16 @@ public class UploadReceiptServlet extends HttpServlet {
     BlobKey blobKey = getUploadedBlobKey(request, "receipt-image");
     long timestamp = clock.instant().toEpochMilli();
     String label = request.getParameter("label");
+    String store = request.getParameter("store");
+    double price = roundPrice(request.getParameter("price"));
 
     // Populate a receipt entity with the information extracted from the image with Cloud Vision.
     Entity receipt = analyzeReceiptImage(blobKey, request);
     receipt.setProperty("blobKey", blobKey);
     receipt.setProperty("timestamp", timestamp);
     receipt.setProperty("label", label);
+    receipt.setProperty("store", store);
+    receipt.setProperty("price", price);
 
     return receipt;
   }
@@ -179,6 +183,13 @@ public class UploadReceiptServlet extends HttpServlet {
   }
 
   /**
+   * Converts a price string into a double rounded to 2 decimal places.
+   */
+  private static double roundPrice(String price) {
+    return Math.round(Double.parseDouble(price) * 100.0) / 100.0;
+  }
+
+  /**
    * Extracts the raw text from the image with the Cloud Vision API. Returns a receipt
    * entity populated with the extracted fields.
    */
@@ -203,15 +214,9 @@ public class UploadReceiptServlet extends HttpServlet {
       throw new ReceiptAnalysisException("Receipt analysis failed.", e);
     }
 
-    // TODO: Replace hard-coded values using receipt analysis with Cloud Vision.
-    double price = 5.89;
-    String store = "McDonald's";
-
     // Create an entity with a kind of Receipt.
     Entity receipt = new Entity("Receipt");
     receipt.setProperty("imageUrl", imageUrl);
-    receipt.setProperty("price", price);
-    receipt.setProperty("store", store);
     // Text objects wrap around a string of unlimited size while strings are limited to 1500 bytes.
     receipt.setUnindexedProperty("rawText", new Text(results.getRawText()));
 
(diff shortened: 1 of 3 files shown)


src/test/java/UploadReceiptServletTest.java
```diff
@@ -205,6 +209,8 @@ public void doPostUploadsReceiptToDatastoreDevServer() throws IOException {
     when(blobInfoFactory.loadBlobInfo(BLOB_KEY)).thenReturn(blobInfo);
 
     when(request.getParameter("label")).thenReturn(LABEL);
```
emilyvera, line 211: I mentioned this in another PR but since a lot of this setup code for each test is repeated, maybe it can be put into a helper function

basaam0: I've added some helper functions here and in PR #27.

src/test/java/UploadReceiptServletTest.java
```diff
@@ -93,6 +93,8 @@
   private static final String LABEL = "Label";
   private static final Text RAW_TEXT = new Text("raw text");
   private static final double PRICE = 5.89;
+  private static final double PRICE_THREE_DECIMAL_PLACES = 17.236;
+  private static final double PRICE_THREE_DECIMAL_PLACES_ROUNDED = 17.24;
```
bryyeh, line 97: If the variable isn't going to be used by more than one test, I prefer to place it inside the test method. This way the list of class variables doesn't become long and unwieldy.

src/test/java/UploadReceiptServletTest.java
```diff
@@ -329,4 +319,60 @@ public void doPostThrowsIfReceiptAnalysisFails() throws IOException {
...
+    Assert.assertEquals(receipt.getProperty("rawText"), RAW_TEXT);
+    Assert.assertEquals(receipt.getProperty("blobKey"), BLOB_KEY);
+    Assert.assertEquals(receipt.getProperty("timestamp"), timestamp);
+    Assert.assertEquals(receipt.getProperty("label"), LABEL);
+  }
+
+  /**
+   * Stubs the request with the given label, store, and price parameters.
+   */
+  private void stubRequestBody(
```
bryyeh, line 361: I like the helper methods :)

basaam0:
> Is there a way to give the user feedback that the total price should only be 2 decimal places in the input box? This way the price isn't unexpectedly rounded for them.

The input box already restricts the user to 2 decimal places. The rounding is meant to be a safeguard on the back end and would not have any effect if the user submits from our front end.

src/main/java/servlets/UploadReceiptServlet.java
```diff
@@ -216,6 +220,13 @@ private static boolean isValidFilename(String filename) {
     return validFilename.matcher(filename).matches();
   }
 
+  /**
+   * Converts a price string into a double rounded to 2 decimal places.
+   */
+  private static double roundPrice(String price) {
+    return Math.round(Double.parseDouble(price) * 100.0) / 100.0;
```
bryyeh, line 227: Do we also need a check to make sure the price is not negative?

src/test/java/UploadReceiptServletTest.java
```diff
@@ -394,6 +381,42 @@ public void doPostThrowsIfReceiptAnalysisFails() throws IOException {
     verify(blobstoreService).delete(BLOB_KEY);
   }
 
+  @Test
+  public void doPostRoundPrice() throws IOException {
+    helper.setEnvIsLoggedIn(true);
+
+    StringWriter stringWriter = new StringWriter();
+    PrintWriter writer = new PrintWriter(stringWriter);
+    when(response.getWriter()).thenReturn(writer);
```
acheng1, line 390: Why is this needed for this test but not for doPostUploadsReceiptToDatastoreLiveServer?

src/test/java/UploadReceiptServletTest.java
```diff
@@ -394,6 +381,42 @@ public void doPostThrowsIfReceiptAnalysisFails() throws IOException {
...
+    PreparedQuery results = datastore.prepare(query);
+    Entity receipt = results.asSingleEntity();
+
+    Assert.assertEquals(receipt.getProperty("imageUrl"), IMAGE_URL);
+    Assert.assertEquals(receipt.getProperty("price"), roundedPrice);
+    Assert.assertEquals(receipt.getProperty("store"), STORE);
+    Assert.assertEquals(receipt.getProperty("rawText"), RAW_TEXT);
+    Assert.assertEquals(receipt.getProperty("blobKey"), BLOB_KEY);
+    Assert.assertEquals(receipt.getProperty("timestamp"), PAST_TIMESTAMP);
+    Assert.assertEquals(receipt.getProperty("label"), LABEL);
```
acheng1, line 417: I would probably only assert on the price since the others are redundant with the assertions in doPostUploadsReceiptToDatastoreLiveServer. Generally we don't want to assert more than is required to verify a behavior.

PR #24, opened 2020-07-10
Add input field for transaction date to upload page.

basaam0:
upload1
upload2

### DIFF
--- a/src/main/webapp/css/upload.css
+++ b/src/main/webapp/css/upload.css
@@ -34,6 +34,10 @@ input {
   width: 7em;
 }
 
+#date-input {
+  width: 12em;
+}
+
 .custom-file {
   font-size: 0.5rem;
 }
@@ -82,7 +86,8 @@ small {
     width: 20em;
   }
 
-  #price-input {
+  #price-input,
+  #date-input {
     font-size: 20px;
   }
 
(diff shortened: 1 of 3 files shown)


src/main/webapp/js/upload.js
```diff
@@ -136,3 +138,24 @@ function checkFileSize() {
...
+ * Sets the value and max value of the transaction date input field to the
+ * current date.
+ */
+function loadDateInput() {
+  const dateInput = document.getElementById('date-input');
+  dateInput.value = dateInput.max = formatDate(new Date());
+}
+
+/**
+ * Converts a date to 'YYYY-MM-DD' format.
```
emilyvera, line 152: Why this format specifically? Just curious

basaam0: This format is required to update the value of the date. The actual UI uses mm/dd/yyyy.

https://www.w3schools.com/Jsref/prop_date_value.asp

src/main/webapp/css/upload.css
```diff
@@ -82,7 +86,8 @@ small {
     width: 20em;
   }
 
-  #price-input {
+  #price-input,
+  #date-input {
```
bryyeh, line 90: I believe these should be sorted alphabetically.

PR #25, opened 2020-07-10
Updated search UI

emilyvera:
- Add new fields for category, store name, and price range.
- Minor fixes on UI to make it fully responsive/prettier :)

Screenshot 2020-07-10 at 10 44 53 AM


src/main/webapp/css/home.css
```diff
@@ -15,3 +16,20 @@
 .error-message {
   padding: 5% 15%;
 }
+
+.search-bar {
+  background-color: #f7f7f7;
+  padding-bottom: 0.8%;
+  padding-top: 0.3%;
```
basaam0, line 23: From the style guide:

> Using shorthand properties is useful for code efficiency and understandability.

https://google.github.io/styleguide/htmlcssguide.html#Shorthand_Properties

src/main/webapp/css/home.css
```diff
@@ -15,3 +16,20 @@
...
+}
+
+.no-padding {
+  padding: 2% 0px 0px 0px !important;
```
basaam0, line 27: > Omit unit specification after “0” values, unless required.

https://google.github.io/styleguide/htmlcssguide.html#0_and_Units

Also, since the left and right padding is the same, the shorthand only needs three values.

https://developer.mozilla.org/en-US/docs/Web/CSS/Shorthand_properties#:~:text=When%20three%20values%20are%20specified,in%20that%20order%20(clockwise).

src/main/webapp/js/index.js
```diff
@@ -127,3 +127,79 @@ $(function() {
...
+      input.addEventListener('keydown', function(e) {
+        const values = keypressSlider.noUiSlider.get();
+        const value = Number(values[handle]);
+        const steps = keypressSlider.noUiSlider.steps();
+        const step = steps[handle];
+        let position;
+
+        // 13 is enter, 38 is key up, 40 is key down.
+        switch (e.which) {
+          case 13:
```
basaam0, line 171: These values can be extracted to constants to make the code more self-documenting

https://drive.google.com/file/d/1uzaIjN8oLqdVL6Sp0VE7ZnBoehEfg_rz/view

src/main/webapp/js/index.js
```diff
@@ -127,3 +127,79 @@ $(function() {
...
+            setSliderHandle(handle, this.value);
+            break;
+          case 38:
+            // Get step to go increase slider value (up)
```
basaam0, line 175: nit: period

src/main/webapp/js/index.js
```diff
@@ -127,3 +127,79 @@ $(function() {
...
+
+  noUiSlider.create(
+      keypressSlider,
+      {start: [20, 80], connect: true, step: 1, range: {min: [0], max: [250]}});
+
+  keypressSlider.noUiSlider.on('update', function(values, handle) {
+    inputs[handle].value = values[handle];
+
+    /* begin Listen to keypress on the input */
+    function setSliderHandle(i, value) {
```
basaam0, line 150: Nested functions should be assigned a name using `const`.

https://google.github.io/styleguide/jsguide.html#features-functions-nested-functions

src/main/webapp/js/index.js
```diff
@@ -127,3 +127,79 @@ $(function() {
...
+      const r = [null, null];
+      r[i] = value;
+      keypressSlider.noUiSlider.set(r);
```
basaam0, line 153: Variables names for `r` and `i` could be more descriptive.

https://google.github.io/styleguide/jsguide.html#naming-rules-common-to-all-identifiers

src/main/webapp/js/index.js
```diff
@@ -127,3 +127,79 @@ $(function() {
...
+    }
+
+    // Listen to keydown events on the input field.
+    inputs.forEach(function(input, handle) {
+      input.addEventListener('change', function() {
+        setSliderHandle(handle, this.value);
+      });
+
+      input.addEventListener('keydown', function(e) {
```
basaam0, line 162: Naming: `event` instead of `e`

src/main/webapp/js/index.js
```diff
@@ -159,34 +159,36 @@ $(document).ready(function() {
...
         const steps = keypressSlider.noUiSlider.steps();
         const step = steps[handle];
         let position;
 
-        // 13 is enter, 38 is key up, 40 is key down.
-        switch (e.which) {
-          case 13:
+        const enter = 13;
+        const up = 38;
+        const down = 40;
```
basaam0, line 171: From the style guide: Constant names use `CONSTANT_CASE`

https://google.github.io/styleguide/jsguide.html#naming-constant-names

acheng1: So the constants above should be `ENTER`, `UP` and `DOWN`?

PR #26, opened 2020-07-10
First unit test for ReceiptAnalysis

avagarcia0:
I wanted to get feedback on this test before writing additional ones, so I'm submitting it as its own pull request.

### DIFF
--- a/src/test/java/ReceiptAnalysisTest.java
+++ b/src/test/java/ReceiptAnalysisTest.java
@@ -27,10 +27,13 @@ import com.google.cloud.vision.v1.Image;
 import com.google.cloud.vision.v1.ImageAnnotatorClient;
 import com.google.common.collect.ImmutableList;
 import com.google.protobuf.ByteString;
+import com.google.sps.data.AnalysisResults;
+import com.google.sps.servlets.ReceiptAnalysis;
 import java.io.ByteArrayInputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.net.URL;
+import org.junit.Assert;
 import org.junit.Before;
 import org.junit.Test;
 import org.junit.runner.RunWith;
@@ -44,7 +47,7 @@ import org.powermock.modules.junit4.PowerMockRunner;
 @RunWith(PowerMockRunner.class)
 @PrepareForTest({AnnotateImageResponse.class, AnnotateImageRequest.class,
     BatchAnnotateImagesResponse.class, ByteString.class, EntityAnnotation.class, Feature.class,
-    Image.class, ImageAnnotatorClient.class, URL.class})
+    Image.class, ImageAnnotatorClient.class, ReceiptAnalysis.class, URL.class})
 public final class ReceiptAnalysisTest {
   private static final ByteString IMAGE_BYTES = ByteString.copyFromUtf8("byte string");
   private static final String RAW_TEXT = "raw text";
@@ -100,5 +103,9 @@ public final class ReceiptAnalysisTest {
     when(response.getTextAnnotationsList()).thenReturn(annotations);
 
     when(annotation.getDescription()).thenReturn(RAW_TEXT);
+
+    AnalysisResults results = ReceiptAnalysis.serveImageText(url);
+
+    Assert.assertEquals(results.getRawText(), RAW_TEXT);
   }
 }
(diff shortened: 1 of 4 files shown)


src/test/java/ReceiptAnalysisTest.java
```diff
@@ -0,0 +1,111 @@
...
+    mockStatic(Image.class);
+    when(Image.newBuilder()).thenReturn(imageBuilder);
+
+    when(featureBuilder.build()).thenReturn(feature);
+    mockStatic(Feature.class);
+    when(Feature.newBuilder()).thenReturn(featureBuilder);
+
+    when(requestBuilder.build()).thenReturn(request);
+    mockStatic(AnnotateImageRequest.class);
+    when(AnnotateImageRequest.newBuilder()).thenReturn(requestBuilder);
```
basaam0, line 87: Reordering these blocks so that the stubbing goes in the same order as the execution of the methods would make it easier to follow.

For example, since `Image.newBuilder()` is called before `build()` in the ReceiptAnalysis class, the corresponding order is:
```
mockStatic(Image.class);
when(Image.newBuilder()).thenReturn(imageBuilder);
when(imageBuilder.build()).thenReturn(image);
```

src/test/java/ReceiptAnalysisTest.java
```diff
@@ -0,0 +1,111 @@
...
+
+    BatchAnnotateImagesResponse batchResponse = mock(BatchAnnotateImagesResponse.class);
+    when(client.batchAnnotateImages(Mockito.<AnnotateImageRequest>anyList()))
+        .thenReturn(batchResponse);
+
+    AnnotateImageResponse response = mock(AnnotateImageResponse.class);
+    ImmutableList<AnnotateImageResponse> responses = ImmutableList.of(response);
+    when(batchResponse.getResponsesList()).thenReturn(responses);
+
+    EntityAnnotation annotation = mock(EntityAnnotation.class);
```
basaam0, line 101: Based on tott/289, I don't think classes like EntityAnnotation or AnnotateImageResponse need to be mocked since they don't have network interaction or file IO.

https://drive.google.com/file/d/0B-yWrS8S9kyqTmFFa1NWVWs5c2c/preview

src/test/java/ReceiptAnalysisTest.java
```diff
@@ -0,0 +1,75 @@
...
+    ImageAnnotatorClient client = mock(ImageAnnotatorClient.class);
+    mockStatic(ImageAnnotatorClient.class);
+    when(ImageAnnotatorClient.create()).thenReturn(client);
+
+    EntityAnnotation annotation = EntityAnnotation.newBuilder().setDescription(RAW_TEXT).build();
+    AnnotateImageResponse response =
+        AnnotateImageResponse.newBuilder().addTextAnnotations(annotation).build();
+    BatchAnnotateImagesResponse batchResponse =
+        BatchAnnotateImagesResponse.newBuilder().addResponses(response).build();
+    when(client.batchAnnotateImages(Mockito.<AnnotateImageRequest>anyList()))
```
bryyeh, line 68: Can you also verify that the input into the api was correct.

src/test/java/ReceiptAnalysisTest.java
```diff
@@ -0,0 +1,75 @@
...
+  private static final ByteString IMAGE_BYTES = ByteString.copyFromUtf8("byte string");
+  private static final String RAW_TEXT = "raw text";
+
+  @Before
+  public void setUp() {
+    MockitoAnnotations.initMocks(this);
+  }
+
+  @Test
+  public void serveImageTextUrlReturnsAnalysisResults() throws IOException {
```
bryyeh, line 54: Are you planning to add more unit tests?

src/test/java/ReceiptAnalysisTest.java
avagarcia0: What's a good way to do this? I know I could build the list in the test, but that would basically just be copying code from the method I'm testing.

src/test/java/ReceiptAnalysisTest.java
avagarcia0: Yes

avagarcia0: PR #31 adds some more

PR #30, opened 2020-07-13
Add set of user-assigned categories to receipt entity.

basaam0:
- Remove label field from receipt entity
- Add category field from request parameters
- Update upload unit tests

### DIFF
--- a/src/main/java/servlets/UploadReceiptServlet.java
+++ b/src/main/java/servlets/UploadReceiptServlet.java
@@ -36,9 +36,11 @@ import java.time.Clock;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.Set;
 import java.util.logging.Logger;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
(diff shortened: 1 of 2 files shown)


src/main/java/servlets/UploadReceiptServlet.java
```diff
@@ -273,6 +280,13 @@ private String getBaseUrl(HttpServletRequest request) {
     return baseUrl;
   }
 
+  /**
+   * Gets the list of user-assigned categories from the request.
+   */
+  private Set<String> getCategories(HttpServletRequest request) {
+    return new HashSet<>(Arrays.asList(request.getParameterValues("categories")));
```
bryyeh, line 287: I believe if there are duplicate categories, this will throw an exception. Is that what we want? If so, we may want to tell the user.

src/main/java/servlets/UploadReceiptServlet.java
```diff
@@ -273,6 +280,13 @@ private String getBaseUrl(HttpServletRequest request) {
     return baseUrl;
   }
 
+  /**
+   * Gets the list of user-assigned categories from the request.
+   */
+  private Set<String> getCategories(HttpServletRequest request) {
```
bryyeh, line 286: Use ImmutableSet.

As Effective Java Item 17 explains, "immutable classes are easier to design, implement, and use than mutable classes. They are less prone to error and are more secure."

src/main/java/servlets/UploadReceiptServlet.java
basaam0: I changed this method to use an ImmutableSet, so now it will remove duplicates. I have also added a unit test in #41 for duplicate categories.

PR #31, opened 2020-07-14
Handle bad response from Cloud Vision API

avagarcia0:
Throw an exception for:
* empty batch response
* response with error
* empty text annotations list

### DIFF
--- a/src/main/java/servlets/ReceiptAnalysis.java
+++ b/src/main/java/servlets/ReceiptAnalysis.java
@@ -42,7 +42,8 @@ import javax.servlet.http.HttpServletResponse;
  */
 public class ReceiptAnalysis {
   /** Returns the text of the image at the requested URL. */
-  public static AnalysisResults serveImageText(URL url) throws IOException {
+  public static AnalysisResults serveImageText(URL url)
+      throws IOException, ReceiptAnalysisException {
     ByteString imageBytes = readImageBytes(url);
 
     String rawText = retrieveText(imageBytes);
@@ -52,7 +53,8 @@ public class ReceiptAnalysis {
   }
 
   /** Returns the text of the image at the requested blob key. */
-  public static AnalysisResults serveImageText(BlobKey blobKey) throws IOException {
+  public static AnalysisResults serveImageText(BlobKey blobKey)
+      throws IOException, ReceiptAnalysisException {
     ByteString imageBytes = readImageBytes(blobKey);
 
     String rawText = retrieveText(imageBytes);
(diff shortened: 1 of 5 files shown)


src/main/java/servlets/ReceiptAnalysis.java
```diff
@@ -108,10 +111,21 @@ private static String retrieveText(ByteString imageBytes) throws IOException {
     ImmutableList<AnnotateImageRequest> requests = ImmutableList.of(request);
 
     try (ImageAnnotatorClient client = ImageAnnotatorClient.create()) {
-      // TODO: Throw custom exception from PR #12 if response has an error or is missing
       BatchAnnotateImagesResponse batchResponse = client.batchAnnotateImages(requests);
```
bryyeh, line 114: Do we need to worry about this api call throwing an exception?

src/test/java/ReceiptAnalysisTest.java
```diff
@@ -83,4 +97,72 @@ public void serveImageTextUrlReturnsAnalysisResults() throws IOException {
...
+    ImageAnnotatorClient client = mock(ImageAnnotatorClient.class);
+    mockStatic(ImageAnnotatorClient.class);
+    when(ImageAnnotatorClient.create()).thenReturn(client);
+
+    BatchAnnotateImagesResponse batchResponse = BatchAnnotateImagesResponse.newBuilder().build();
+    when(client.batchAnnotateImages(Mockito.<AnnotateImageRequest>anyList()))
+        .thenReturn(batchResponse);
+
+    expectedException.expect(ReceiptAnalysisException.class);
+    expectedException.expectMessage(EMPTY_BATCH_RESPONSE_WARNING);
```
bryyeh, line 117: For testing exceptions, I recommend using assertThrows(). You want to break your test up into three parts: arrange, act, and assert. You may need to upgrade your version of junit for this.

https://stackoverflow.com/questions/40268446/junit-5-how-to-assert-an-exception-is-thrown
https://howtodoinjava.com/junit5/expected-exception-example/

src/test/java/ReceiptAnalysisTest.java
```diff
@@ -160,9 +162,31 @@ public void serveImageTextThrowsIfEmptyTextAnnotationsList()
...
+
+  @Test
+  public void serveImageTextThrowsIfRequestFails() throws IOException, ReceiptAnalysisException {
+    URL url = mock(URL.class);
+    InputStream inputStream = new ByteArrayInputStream(IMAGE_BYTES.toByteArray());
+    when(url.openStream()).thenReturn(inputStream);
+
+    ImageAnnotatorClient client = mock(ImageAnnotatorClient.class);
+    mockStatic(ImageAnnotatorClient.class);
+    when(ImageAnnotatorClient.create()).thenReturn(client);
```
bryyeh, line 179: If these are the same in all your tests, you may want to put it in you setUp() method.

PR #32, opened 2020-07-14
Refactor: remove label and add categories

emilyvera:
Label was only used for the prototype, so I removed that across all files and replaced it with categories instead

### DIFF
--- a/src/main/java/servlets/SearchServlet.java
+++ b/src/main/java/servlets/SearchServlet.java
@@ -80,4 +80,3 @@ public class SearchServlet extends HttpServlet {
     return new Receipt(id, userId, timestamp, blobKey, imageUrl, price, store, categories, rawText);
   }
 }
-}
--- a/src/main/java/servlets/SearchServlet.java
+++ b/src/main/java/servlets/SearchServlet.java
@@ -20,14 +20,17 @@ import com.google.appengine.api.datastore.DatastoreServiceFactory;
 import com.google.appengine.api.datastore.Entity;
 import com.google.appengine.api.datastore.FetchOptions;
 import com.google.appengine.api.datastore.Query;
-import com.google.appengine.api.datastore.Query.Filter;
 import com.google.appengine.api.datastore.Query.FilterOperator;
-import com.google.appengine.api.datastore.Query.FilterPredicate;
+import com.google.appengine.api.users.UserService;
+import com.google.appengine.api.users.UserServiceFactory;
+import com.google.common.base.Strings;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
 import com.google.gson.Gson;
+import com.google.sps.data.QueryInformation;
 import com.google.sps.data.Receipt;
 import java.io.IOException;
+import java.text.ParseException;
 import java.util.ArrayList;
 import java.util.stream.Stream;
 import javax.servlet.annotation.WebServlet;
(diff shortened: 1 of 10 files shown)


src/main/webapp/js/index.js
```diff
@@ -15,12 +15,12 @@
 
 /** Fetches receipts from the server and adds them to the DOM. */
 async function searchReceipts() {
-  const label = document.getElementById('search-input').value;
-  const response = await fetch(`/search-receipts?label=${label}`);
+  const categories = document.getElementById('category-input').value;
+  const response = await fetch(`/search-receipts?categories=${categories}`);
```
acheng1, line 19: Similar comment to your other PR: I would consider using [URLSearchParams](https://developers.google.com/web/updates/2016/01/urlsearchparams) to build the parameter list or use encodeURIComponent to make sure the user input is properly encoded (e.g., will the above won't handle if the input includes a space like "burger king").

src/main/java/servlets/SearchServlet.java
```diff
@@ -47,20 +40,20 @@
...
     response.getWriter().println(gson.toJson(receipts));
   }
 
-  /** Returns ImmutableList of receipts from datastore with the same label as desiredLabel. */
-  private ImmutableList<Receipt> getReceiptsWithMatchingLabel(String desiredLabel) {
-    // Set filter to retrieve only receipts with label equal to desiredLabel.
-    Filter matchingLabels = new FilterPredicate("label", FilterOperator.EQUAL, desiredLabel);
+  /** Returns ImmutableList of receipts from datastore with the same categories. */
+  private ImmutableList<Receipt> getReceiptsWithMatchingCategories(String categories) {
+    // Set filter to retrieve only receipts with categories equal to categories.
```
acheng1, line 53: Does this mean the user has to remember the exact set of categories they originally entered or that was extracted from the receipt? I would have assumed that if we had the following:

receipt1 - category1, category2, category3
receipt2 - category2
receipt3 - category3
receipt4 - category4

searching for category1 would return receipt1 and searching for category2 would return receipt1 and receipt2. If you searched for category3, category4, the search would return receipt1, receipt3, and receipt4 so would be an OR of the categories the user is searching for.

emilyvera: Searching for only one category would return the right receipts. But for the MVP we agreed that the user would only be able to search one category for now

src/main/webapp/js/index.js
emilyvera: It handles spaces already

src/main/java/servlets/SearchServlet.java
```diff
@@ -47,20 +40,20 @@
...
     response.setContentType("application/json;");
     response.getWriter().println(gson.toJson(receipts));
   }
 
-  /** Returns ImmutableList of receipts from datastore with the same label as desiredLabel. */
-  private ImmutableList<Receipt> getReceiptsWithMatchingLabel(String desiredLabel) {
-    // Set filter to retrieve only receipts with label equal to desiredLabel.
-    Filter matchingLabels = new FilterPredicate("label", FilterOperator.EQUAL, desiredLabel);
+  /** Returns ImmutableList of receipts from datastore with the same categories. */
+  private ImmutableList<Receipt> getReceiptsWithMatchingCategories(String categories) {
```
bryyeh, line 52: If only one category can be searched for this version, can you change all the plural "categories" to be singular.

acheng1:
> Label was only used for the prototype, so I removed that across all files and replaced it with categories instead

Thanks for cleaning this up!

PR #33, opened 2020-07-14
Retrieve query from form and attach to query string

emilyvera:


src/main/webapp/js/index.js
```diff
@@ -16,7 +16,16 @@
...
+  const dateRange = document.getElementById('date-range-input').textContent;
+  const store = document.getElementById('store-name-input').value;
+  const minPrice = document.getElementById('min-price-input').value;
+  const maxPrice = document.getElementById('max-price-input').value;
+  const dateTimeFormat = new Intl.DateTimeFormat();
+  const timeZoneId = dateTimeFormat.resolvedOptions().timeZone;
+
+  const response = await fetch(`/search-receipts?categories=
+    ${categories}&store=${store}&min=${minPrice}&max=
+    ${maxPrice}&dateRange=${dateRange}&timeZoneId=${timeZoneId}`);
```
acheng1, line 28: I would consider using [URLSearchParams](https://developers.google.com/web/updates/2016/01/urlsearchparams) to build the parameter list or use encodeURIComponent to make sure the user input is properly encoded (e.g., contacting the string above won't handle if the input includes a space like "burger king").

PR #34, opened 2020-07-14
Add query information class

emilyvera:
QueryInformation will hold the query fields retrieved from the form


src/main/java/data/QueryInformation.java
```diff
@@ -0,0 +1,104 @@
...
+import com.google.common.base.Strings;
+import com.google.common.collect.ImmutableSet;
+import java.text.DateFormat;
+import java.text.ParseException;
+import java.util.Arrays;
+import java.util.TimeZone;
+
+/** Class to hold query information. */
+public class QueryInformation {
+  /** Millieconds equivalent to 11:59:59 PM. */
```
basaam0, line 26: typo: "Milliseconds"

src/main/java/data/QueryInformation.java
```diff
@@ -0,0 +1,104 @@
...
+  private static final long MILLISECONDS_TO_END_OF_DAY = (24L * 60L * 60L * 1000L) - 1000L;
```
basaam0, line 27: nit: Would it change the results if only 1 ms was subtracted instead of 1000? I know right now the users can't specify any unit of time smaller than a day when uploading a receipt, but in case a receipt timestamp was at 11:59:59.999, it would technically be in the same day and should be returned in the query.

src/main/java/data/QueryInformation.java
```diff
@@ -0,0 +1,104 @@
...
+
+  public QueryInformation(String timeZoneId, String categories, String dateRange, String store,
+      String minPrice, String maxPrice)
+      throws ParseException, NumberFormatException, NullPointerException {
+    this.timeZone = TimeZone.getTimeZone(timeZoneId);
+
+    String formattedCategories;
+    try {
+      formattedCategories = formatInput(categories);
+    } catch (NullPointerException exception) {
```
bryyeh, line 45: Are we worried about formatInput() throwing an exception because "categories" might be null? Is that an expected input to QueryInformation?

src/test/java/QueryInformationTest.java
```diff
@@ -0,0 +1,146 @@
...
+    // Test with empty string.
+    QueryInformation queryInformation = new QueryInformation(
+        CST_TIMEZONE_ID, EMPTY_STRING, DATE_RANGE, STORE, MIN_PRICE, MAX_PRICE);
+    Assert.assertEquals(EMPTY_IMMUTABLE_SET, queryInformation.getCategories());
+
+    // Test with null.
+    queryInformation =
+        new QueryInformation(CST_TIMEZONE_ID, NULL_VALUE, DATE_RANGE, STORE, MIN_PRICE, MAX_PRICE);
+    Assert.assertEquals(EMPTY_IMMUTABLE_SET, queryInformation.getCategories());
+  }
```
bryyeh, line 97: I would split these into 2 tests.

src/test/java/QueryInformationTest.java
```diff
@@ -0,0 +1,146 @@
+// Copyright 2019 Google LLC
```
bryyeh, line 1: Tests are looking pretty nice :)

src/test/java/QueryInformationTest.java
```diff
@@ -0,0 +1,146 @@
...
+import org.junit.Assert;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+@RunWith(JUnit4.class)
+public final class QueryInformationTest {
+  private static final String EMPTY_STRING = "";
+  private static final String NULL_VALUE = null;
+  private static final ImmutableSet<String> EMPTY_IMMUTABLE_SET = ImmutableSet.of();
```
bryyeh, line 30: nitnit: not sure these are necessary. You could probably reference them directly in the test.

src/test/java/QueryInformationTest.java
```diff
@@ -0,0 +1,146 @@
...
+    QueryInformation queryInformation = new QueryInformation(
+        CST_TIMEZONE_ID, EDGE_CASE_CATEGORIES, DATE_RANGE, STORE, MIN_PRICE, MAX_PRICE);
+    Assert.assertEquals(EXPECTED_CATEGORIES, queryInformation.getCategories());
+  }
+
+  @Test
+  public void invalidCategories() throws ParseException {
+    // Test with empty string.
+    QueryInformation queryInformation = new QueryInformation(
+        CST_TIMEZONE_ID, EMPTY_STRING, DATE_RANGE, STORE, MIN_PRICE, MAX_PRICE);
```
bryyeh, line 90: If parameter isn't self-explanatory, add a comment before it. Example:
QueryInformation(CST_TIMEZONE_ID, /* categories = */ EMPTY_STRING)

src/test/java/QueryInformationTest.java
```diff
@@ -0,0 +1,146 @@
...
+
+  @Test
+  public void pricesCorrectlySet() throws ParseException {
+    QueryInformation queryInformation =
+        new QueryInformation(CST_TIMEZONE_ID, CATEGORIES, DATE_RANGE, STORE, MIN_PRICE, MAX_PRICE);
+    Assert.assertEquals(EXPECTED_MIN_PRICE, queryInformation.getMinPrice(), ERROR_THRESHOLD);
+    Assert.assertEquals(EXPECTED_MAX_PRICE, queryInformation.getMaxPrice(), ERROR_THRESHOLD);
+  }
+
+  @Test(expected = NumberFormatException.class)
```
bryyeh, line 135: I asked Ava to do this too, can we use assertThrows(). We want to break your test up into three parts: arrange, act, and assert. You may need to upgrade your version of junit for this.

https://stackoverflow.com/questions/40268446/junit-5-how-to-assert-an-exception-is-thrown
https://howtodoinjava.com/junit5/expected-exception-example/
https://github.com/googleinterns/step12-2020/pull/31/commits/de5cd78b6d147bb98a7e05a9a5da49e7a586678d

src/main/java/data/QueryInformation.java
emilyvera: Since I built the front end and back end, I know categories will never be null. At the worst, it'll just be an empty string. However, I remember getting feedback that we should build the back end with the mindset that someone else is building the front and you don't know what kinds of inputs you'll get from them. That's why I did this here

src/test/java/QueryInformationTest.java
emilyvera: Thanks!

src/main/java/data/QueryInformation.java
acheng1: I think maybe @bryyeh is suggesting that the request parameter validation could be done in SearchServlet.java prior to calling this class. We generally validate the parameters as soon as we can so we can fail fast.

src/main/java/data/QueryInformation.java
```diff
@@ -0,0 +1,104 @@
...
+      formattedCategories = null;
+    }
+
+    if (!Strings.isNullOrEmpty(formattedCategories)) {
```
acheng1, line 49: I'd rearrange this as the following so that the else clause doesn't read as not not null or empty:
```
if (Strings.isNullOrEmpty(...)) {
  this.categories = ImmutableSet.of();
} else {
  ...
}
```
Otherwise

src/main/java/data/QueryInformation.java
emilyvera: Gotcha, well in that case, I removed the try/catch statement here. The servlet code in #37 in createQueryInformation() handles nulls so you're right that they shouldn't be an expected input to QueryInformation

PR #35, opened 2020-07-14
Redirect to receipt analysis page after uploading receipt.

basaam0:
- Instead of redirecting to the home page after a receipt is successfully uploaded, redirect to the receipt analysis page
- Send URL parameters containing information about the newly uploaded receipt so that the receipt analysis page can display it

### DIFF
--- a/src/main/webapp/js/upload.js
+++ b/src/main/webapp/js/upload.js
@@ -53,10 +53,15 @@ async function uploadReceipt(event) {
   }
 
   const json = (await response.json()).propertyMap;
+  const params = new URLSearchParams();
+  params.append('categories', json.categories);
+  params.append('image-url', json.imageUrl);
+  params.append('price', json.price);
+  params.append('store', json.store);
+  params.append('timestamp', json.timestamp);
+
   // Redirect to the receipt analysis page.
-  window.location.href = `/receipt-analysis.html?categories=` +
-      `${json.categories}&image-url=${json.imageUrl}` +
-      `&price=${json.price}&store=${json.store}&timestamp=${json.timestamp}`;
+  window.location.href = `/receipt-analysis.html?${params.toString()}`;
 }
 
 /**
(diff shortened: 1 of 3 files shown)


src/main/webapp/js/upload.js
```diff
@@ -51,8 +52,11 @@ async function uploadReceipt(event) {
     return;
   }
 
-  // TODO: Redirect to receipt analysis page for MVP
-  window.location.href = '/';
+  const json = (await response.json()).propertyMap;
+  // Redirect to the receipt analysis page.
+  window.location.href = `/receipt-analysis.html?categories=` +
+      `${json.categories}&image-url=${json.imageUrl}` +
+      `&price=${json.price}&store=${json.store}&timestamp=${json.timestamp}`;
```
avagarcia0, line 59: From https://github.com/googleinterns/step12-2020/pull/33#discussion_r454842800:
> I would consider using [URLSearchParams](https://developers.google.com/web/updates/2016/01/urlsearchparams) to build the parameter list or use encodeURIComponent to make sure the user input is properly encoded (e.g., will the above handle if the input includes a space like "burger king").

src/test/java/UploadReceiptServletTest.java
```diff
@@ -214,11 +219,21 @@ public void doPostUploadsReceiptToDatastoreLiveServer() throws IOException {
...
+    String expectedResponse = createReceiptEntity(IMAGE_URL, PRICE, STORE, RAW_TEXT, BLOB_KEY,
+        PAST_TIMESTAMP, CATEGORIES_COLLECTION, USER_ID);
+    Assert.assertTrue(response.contains(expectedResponse));
   }
 
   @Test
   public void doPostUploadsReceiptToDatastoreDevServer() throws IOException {
     helper.setEnvIsLoggedIn(true);
+
+    StringWriter stringWriter = new StringWriter();
```
emilyvera, line 233: Looks like this setup is done in every test, so I would put this setup in the @Before method

PR #36, opened 2020-07-14
Display receipt analysis for newly uploaded receipt

avagarcia0:
* Read URL parameters and extract information about the receipt
* Remove hardcoded values and display parameterized values instead

image

### DIFF
--- a/src/main/webapp/js/receipt-analysis.js
+++ b/src/main/webapp/js/receipt-analysis.js
@@ -16,7 +16,9 @@
 async function loadReceiptAnalysis() {
   const timestamp = new URLSearchParams(location.search).get('timestamp');
   const storeName = new URLSearchParams(location.search).get('store-name');
+  const total = new URLSearchParams(location.search).get('total');
 
   document.getElementById('timestamp').innerText = `Upload Date: ${timestamp}`;
   document.getElementById('store-name').innerText = `Store Name: ${storeName}`;
+  document.getElementById('total').innerText = `Total Price: $${total}`;
 }
--- a/src/main/webapp/js/receipt-analysis.js
+++ b/src/main/webapp/js/receipt-analysis.js
@@ -19,6 +19,7 @@ function loadReceiptAnalysis() {
   const total = new URLSearchParams(location.search).get('total');
   const categories =
       new URLSearchParams(location.search).get('categories').split(',');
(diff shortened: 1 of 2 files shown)


src/main/webapp/js/receipt-analysis.js
```diff
@@ -0,0 +1,33 @@
...
+// limitations under the License.
+
+/** Fetches receipt properties from the server and adds them to the page. */
+function loadReceiptAnalysis() {
+  const timestamp = new URLSearchParams(location.search).get('timestamp');
+  const storeName = new URLSearchParams(location.search).get('store');
+  const total = new URLSearchParams(location.search).get('price');
+  const categories =
+      new URLSearchParams(location.search).get('categories').split(',');
+  const imageUrl = new URLSearchParams(location.search).get('image-url');
```
basaam0, line 22: Adding an intermediate variable for the URL parameters would make this simpler.

src/main/webapp/js/receipt-analysis.js
```diff
@@ -0,0 +1,33 @@
...
+
+  document.getElementById('timestamp').innerText = `Upload Date: ${timestamp}`;
```
basaam0, line 24: In #35, the timestamp URL parameter is in milliseconds, so eventually it should be formatted as a date. The `toLocaleString()` method may help with this.

`const date = new Date(timestamp).toLocaleString('en-US;, {...})`

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleString

src/main/webapp/js/receipt-analysis.js
```diff
@@ -0,0 +1,33 @@
...
+  document.getElementById('store-name').innerText = `Store Name: ${storeName}`;
+  document.getElementById('total').innerText = `Total Price: $${total}`;
```
emilyvera, line 26: Is there supposed to be two $$ here

avagarcia0: Yeah, the first one is the currency symbol. Would it be better to use `'Total Price: $' + total` instead?

emilyvera: Ohhh I see, nope that makes sense now :)

PR #37, opened 2020-07-14
Updated search servlet

emilyvera:
Updated search method:

- Can now search by category, date range, store, and price range. Note: date range and price range are always included in the query, but category and store are optional.
- Includes sanitization of strings from the form (i.e. strings with trailing/leading whitespace and casing)

### DIFF
--- a/src/main/java/data/QueryInformation.java
+++ b/src/main/java/data/QueryInformation.java
@@ -1,0 +1,104 @@
+// Copyright 2019 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.sps.data;
+
+import com.google.common.base.Strings;
+import com.google.common.collect.ImmutableSet;
+import java.text.DateFormat;
+import java.text.ParseException;
+import java.util.Arrays;
+import java.util.TimeZone;
+
+/** Class to hold query information. */
+public class QueryInformation {
+  /** Millieconds equivalent to 11:59:59 PM. */
+  private static final long MILLISECONDS_TO_END_OF_DAY = (24L * 60L * 60L * 1000L) - 1000L;
+
+  private final TimeZone timeZone;
+  private final ImmutableSet<String> categories;
+  private final long startTimestamp;
+  private final long endTimestamp;
+  private final String store;
+  private final double minPrice;
+  private final double maxPrice;
+
+  public QueryInformation(String timeZoneId, String categories, String dateRange, String store,
+      String minPrice, String maxPrice)
+      throws ParseException, NumberFormatException, NullPointerException {
+    this.timeZone = TimeZone.getTimeZone(timeZoneId);
+
+    String formattedCategories;
+    try {
+      formattedCategories = formatInput(categories);
+    } catch (NullPointerException exception) {
+      formattedCategories = null;
+    }
+
+    if (!Strings.isNullOrEmpty(formattedCategories)) {
+      this.categories = ImmutableSet.copyOf(Arrays.asList(formattedCategories));
+    } else {
+      this.categories = ImmutableSet.of(); // Empty set.
+    }
+
+    String[] dates = dateRange.split("-");
+    this.startTimestamp = dateToMilliseconds(formatInput(dates[0]));
+    this.endTimestamp = dateToMilliseconds(formatInput(dates[1])) + MILLISECONDS_TO_END_OF_DAY;
+
+    this.store = formatInput(store);
+    this.minPrice = Double.parseDouble(formatInput(minPrice));
+    this.maxPrice = Double.parseDouble(formatInput(maxPrice));
+  }
+
+  /** Sets input to lowercase and replaces all extra whitespace before/after/between. */
+  private String formatInput(String rawInput) {
+    return rawInput.toLowerCase().replaceAll("\\s+", " ").trim();
+  }
+
+  /** Converts a formatted date (month day, year) to milliseconds since epoch. */
+  private long dateToMilliseconds(String date) throws ParseException {
+    DateFormat dateFormat = DateFormat.getDateInstance(DateFormat.LONG);
+    dateFormat.setTimeZone(timeZone);
+
+    return dateFormat.parse(date).getTime();
+  }
+
+  public TimeZone getTimeZone() {
+    return timeZone;
+  }
+
+  public ImmutableSet<String> getCategories() {
+    return categories;
+  }
+
+  public long getStartTimestamp() {
+    return startTimestamp;
+  }
+
+  public long getEndTimestamp() {
+    return endTimestamp;
+  }
+
+  public String getStore() {
+    return store;
+  }
+
+  public double getMinPrice() {
+    return minPrice;
+  }
+
+  public double getMaxPrice() {
+    return maxPrice;
+  }
+}
--- a/src/main/java/data/QueryInformation.java
+++ b/src/main/java/data/QueryInformation.java
@@ -27,29 +27,29 @@ public class QueryInformation {
   private static final long MILLISECONDS_TO_END_OF_DAY = (24L * 60L * 60L * 1000L) - 1L;
 
   private final TimeZone timeZone;
-  private final ImmutableSet<String> categories;
+  private final ImmutableSet<String> category;
   private final long startTimestamp;
   private final long endTimestamp;
   private final String store;
   private final double minPrice;
   private final double maxPrice;
 
-  public QueryInformation(String timeZoneId, String categories, String dateRange, String store,
+  public QueryInformation(String timeZoneId, String category, String dateRange, String store,
       String minPrice, String maxPrice)
       throws ParseException, NumberFormatException, NullPointerException {
     this.timeZone = TimeZone.getTimeZone(timeZoneId);
 
-    String formattedCategories;
+    String formattedCategory;
     try {
-      formattedCategories = formatInput(categories);
+      formattedCategory = formatInput(category);
     } catch (NullPointerException exception) {
-      formattedCategories = null;
+      formattedCategory = null;
     }
 
-    if (!Strings.isNullOrEmpty(formattedCategories)) {
-      this.categories = ImmutableSet.copyOf(Arrays.asList(formattedCategories));
+    if (!Strings.isNullOrEmpty(formattedCategory)) {
+      this.category = ImmutableSet.copyOf(Arrays.asList(formattedCategory));
     } else {
-      this.categories = ImmutableSet.of(); // Empty set.
+      this.category = ImmutableSet.of(); // Empty set.
     }
 
     String[] dates = dateRange.split("-");
@@ -78,8 +78,8 @@ public class QueryInformation {
     return timeZone;
   }
 
-  public ImmutableSet<String> getCategories() {
-    return categories;
+  public ImmutableSet<String> getCategory() {
+    return category;
   }
 
   public long getStartTimestamp() {
--- a/src/main/java/data/QueryInformation.java
+++ b/src/main/java/data/QueryInformation.java
@@ -23,7 +23,7 @@ import java.util.TimeZone;
 
 /** Class to hold query information. */
 public class QueryInformation {
-  /** Milliseconds equivalent to 11:59:59 PM. */
+  /** Milliseconds equivalent to 11:59:59.999 PM. */
   private static final long MILLISECONDS_TO_END_OF_DAY = (24L * 60L * 60L * 1000L) - 1L;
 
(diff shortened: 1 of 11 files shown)


src/main/java/servlets/SearchServlet.java
```diff
@@ -38,33 +39,102 @@
...
 
-  /** Returns ImmutableList of receipts from datastore with the same categories. */
-  private ImmutableList<Receipt> getReceiptsWithMatchingCategories(String categories) {
-    // Set filter to retrieve only receipts with categories equal to categories.
-    Filter matchingCategories = new FilterPredicate("categories", FilterOperator.EQUAL, categories);
-    Query query = new Query("Receipt");
-    query.setFilter(matchingCategories);
+  /** Creates a {@link QueryInformation} based on request parameters. */
+  private QueryInformation createQueryInformation(
+      HttpServletRequest request, HttpServletResponse response) throws IOException {
```
basaam0, line 77: Not mandatory but this would be a good place to use the Optional container type as a return type.

https://www.oracle.com/technical-resources/articles/java/java8-optional.html
https://docs.oracle.com/javase/8/docs/api/java/util/Optional.html

src/test/java/TestUtils.java
```diff
@@ -0,0 +1,90 @@
...
+    receiptEntity.setProperty("price", price);
+    receiptEntity.setProperty("store", store);
+    receiptEntity.setProperty("categories", categories);
+    receiptEntity.setProperty("rawText", rawText);
+
+    Key key = datastore.put(receiptEntity);
+    return key.getId();
+  }
+
+  /** Set all necessary paremeters that SearchServlet will ask for in a doGet. */
```
basaam0, line 80: typo: "parameters"

src/main/java/servlets/SearchServlet.java
emilyvera: This looks helpful! I'll try to get back to it if I have time

src/main/java/servlets/SearchServlet.java
```diff
@@ -38,33 +39,102 @@
...
+      response.getWriter().println(NULL_EXCEPTION_MESSAGE);
+    } catch (NumberFormatException exception) {
+      response.setStatus(HttpServletResponse.SC_BAD_REQUEST);
+      response.getWriter().println(NUMBER_EXCEPTION_MESSAGE);
+    } catch (ParseException exception) {
+      response.setStatus(HttpServletResponse.SC_BAD_REQUEST);
+      response.getWriter().println(PARSE_EXCEPTION_MESSAGE);
+    }
+
+    return queryInformation;
```
bryyeh, line 101: Instead of returning null when there is an exception, I would let this method propagate the exception back up and have error handling in doGet()

src/main/java/servlets/SearchServlet.java
```diff
@@ -38,33 +39,102 @@
...
+    if (queryInformation.getCategories() != null && queryInformation.getCategories().size() != 0) {
+      query.addFilter("categories", Query.FilterOperator.IN, queryInformation.getCategories());
+    }
+
+    query.addFilter("timestamp", Query.FilterOperator.GREATER_THAN_OR_EQUAL,
+        queryInformation.getStartTimestamp());
+    query.addFilter(
+        "timestamp", Query.FilterOperator.LESS_THAN_OR_EQUAL, queryInformation.getEndTimestamp());
+
+    if (!Strings.isNullOrEmpty(queryInformation.getStore())) {
```
bryyeh, line 131: Today I learned about isNullOrEmpty() :D

src/main/java/servlets/SearchServlet.java
```diff
@@ -38,33 +39,102 @@
...
+  /** Returns ImmutableList of receipts from datastore matching queryInformation fields. */
+  private ImmutableList<Receipt> getMatchingReceipts(QueryInformation queryInformation) {
+    Query query = setupQuery(queryInformation);
 
-    // Convert matching receipt entities to receipt objects and add to return list.
     return datastore.prepare(query)
         .asList(FetchOptions.Builder.withDefaults())
         .stream()
         .map(this::createReceiptFromEntity)
+        .filter(receipt
```
bryyeh, line 112: I would leave a comment about why this couldn't be done as a query filter.

src/test/java/TestUtils.java
```diff
@@ -0,0 +1,90 @@
...
+    receiptEntity.setProperty("store", store);
+    receiptEntity.setProperty("categories", categories);
+    receiptEntity.setProperty("rawText", rawText);
+
+    Key key = datastore.put(receiptEntity);
+    return key.getId();
+  }
+
+  /** Set all necessary parameters that SearchServlet will ask for in a doGet. */
+  public static void setRequestParameters(HttpServletRequest request, String timeZoneId,
```
bryyeh, line 81: nit: setSearchServletRequestParameters

src/test/java/SearchServletTest.java
```diff
@@ -0,0 +1,164 @@
...
+  }
+
+  @After
+  public void tearDown() {
+    helper.tearDown();
+  }
+
+  @Test
+  public void queryWithAllFieldsFilled() throws IOException {
+    // Add mock receipts to datastore.
```
bryyeh, line 91: Can we add a comment section explaining the setup of this test. Similar to the comments in the TDD unit tests. It's difficult to understand the test because there are multiple receipts and search parameters. For example, could you explain why the walmart receipt was returned.

It would be great if this could be done for all these unit tests. Let me know if you have any questions.

src/test/java/TestUtils.java
```diff
@@ -0,0 +1,90 @@
...
+  /** Private constructor to prevent instantiation. */
+  private TestUtils() {
+    throw new UnsupportedOperationException();
+  }
+
+  /* Add receipts to database for testing purposes. */
+  public static void addTestReceipts(DatastoreService datastore) {
+    addTestReceipt(datastore, 1, 123, 1045237591000L, new BlobKey("test"),
+        "img/walmart-receipt.jpg", 26.12, "walmart",
+        new HashSet<>(Arrays.asList("candy", "drink", "personal")), "");
```
bryyeh, line 40: ImmutableSet.of() would be clearer. Same for the rest.

src/test/java/SearchServletTest.java
```diff
@@ -0,0 +1,164 @@
...
+    TestUtils.addTestReceipts(datastore);
```
bryyeh, line 92: Instead of inserting a bunch of receipts for each test, I would insert the minimum amount of receipts needed to test each behavior.

src/test/java/SearchServletTest.java
```diff
@@ -0,0 +1,164 @@
...
+    when(response.getWriter()).thenReturn(writer);
+  }
+
+  @After
+  public void tearDown() {
+    helper.tearDown();
+  }
+
+  @Test
+  public void queryWithAllFieldsFilled() throws IOException {
```
bryyeh, line 90: Can we add a unit test for each search parameter/field.

src/test/java/TestUtils.java
```diff
@@ -0,0 +1,90 @@
...
+
+    addTestReceipt(datastore, 2, 123, 1513103400000L, new BlobKey("test"), "img/canes-receipt.jpg",
+        32.38, "raising cane's chicken fingers",
+        new HashSet<>(Arrays.asList("chicken", "drink", "lunch")), "");
+
+    addTestReceipt(datastore, 3, 123, 1560193140000L, new BlobKey("test"),
```
bryyeh, line 46: For the parameters that aren't self-explanatory such as the ints, can you add a comment before each parameter. Example:
`addTestReceipt(datastore, /* id = */ 3, /* userId = */ 123, /* timestamp = */ 1560193140000L`)

src/test/java/SearchServletTest.java
```diff
@@ -0,0 +1,164 @@
...
+    TestUtils.addTestReceipts(datastore);
+
+    // Perform doGet - this should retrieve one receipt.
+    TestUtils.setRequestParameters(
+        request, CST_TIMEZONE_ID, CATEGORIES, SHORT_DATE_RANGE, STORE, MIN_PRICE, MAX_PRICE);
+    servlet.doGet(request, response);
+    writer.flush();
+
+    // Make sure receipt is retrieved by finding the store name in the writer.
+    Assert.assertTrue(stringWriter.toString().contains("walmart"));
```
bryyeh, line 101: I wonder if Id would be a better field to assert on.

src/test/java/SearchServletTest.java
emilyvera: How would you describe "minimum"? I wanted to show that querying wouldn't just return all the receipts in the database but rather specific ones. So if I expect one receipt to be returned, would adding 2 or 3 in total be sufficient rather than the 6?

src/test/java/SearchServletTest.java
emilyvera: I figured that since there's only one receipt with that store name in these tests, it would be sufficient. But if you think ID works better I can do that

src/test/java/SearchServletTest.java
emilyvera: Yeah, I wasn't sure how many tests would be overdoing it since there's so many different combinations of queries. But this makes sense

emilyvera: Price and timestamp are always included in the query (since on the UI, there's no way for the user to query with an empty date range or price, it will always either be the default values or the values set by the user). So all possible queries include:

1. category, dateRange, store, minPrice, maxPrice
2. dateRange, store, minPrice, maxPrice
3. category, dateRange, minPrice, maxPrice
4. dateRange, minPrice, maxPrice

I originally tested 1 and 4. I've added tests for 2 and 3, but since the user won't only be able to query by each parameter individually (i.e. only by store or only by price), I didn't include tests for those.

src/test/java/SearchServletTest.java
```diff
@@ -88,43 +88,116 @@ public void tearDown() {
...
   public void queryWithAllFieldsFilled() throws IOException {
+    // Columns ommitted from database visual: id, userId, blobKey, imageUrl, rawText.
+    //
+    // id   Timestamp      Price          Store                    Categories
+    // 1  1045237591000    26.12        "walmart"         ["candy", "drink", "personal"]
+    // 2  1560193140000    14.51        "contoso"         ["cappuccino", "sandwich", "lunch"]
+    // 3  1491582960000    29.01   "main st restaurant"   ["food", "meal", "lunch"]
+    //
+    // Query: drink, 2/1/03-2/28/03, walmart, $5.00-$30.00.
+    // Will only return walmart receipt.
```
bryyeh, line 99: Looks amazing!! Thank you for putting this in.

src/test/java/SearchServletTest.java
```diff
@@ -88,43 +88,116 @@ public void tearDown() {
...
+    // Perform doGet - this should retrieve a couple receipts.
+    TestUtils.setSearchServletRequestParameters(
+        request, CST_TIMEZONE_ID, CATEGORY, SHORT_DATE_RANGE, EMPTY_STRING, MIN_PRICE, MAX_PRICE);
     servlet.doGet(request, response);
     writer.flush();
 
-    // Make sure receipt is retrieved by finding the store name in the writer.
-    Assert.assertTrue(stringWriter.toString().contains("walmart"));
+    // Make sure receipts are retrieved by finding receipt ids in the writer.
+    Assert.assertTrue(stringWriter.toString().contains("\"id\":1"));
```
bryyeh, line 136: Can you assert on the other receipts that should be returned? Same for the others.

src/test/java/SearchServletTest.java
```diff
@@ -88,43 +88,116 @@ public void tearDown() {
...
+    // 3  1491582960000    29.01   "main st restaurant"   ["food", "meal", "lunch"]
+    //
+    // Query: 1/1/10-7/31/20, $5.00-$30.00.
+    // Will return main st restaurant and contoso receipts.
+
+    // Add mock receipts to datastore.
+    TestUtils.addTestReceipts(datastore);
+
+    // Perform doGet - this should retrieve a couple receipts.
+    TestUtils.setSearchServletRequestParameters(request, CST_TIMEZONE_ID, EMPTY_STRING,
```
bryyeh, line 180: For the EMPTY_STRING parameters, can you add a comment before it.

src/test/java/TestUtils.java
```diff
@@ -20,49 +20,39 @@
...
+import com.google.common.collect.ImmutableSet;
 import java.io.IOException;
 import java.util.Arrays;
-import java.util.HashSet;
-import java.util.Set;
 import javax.servlet.http.HttpServletRequest;
 
 /** Class that contains helpful methods used for testing. */
 public final class TestUtils {
+  //   public static Map<String, Long> storeIds = new HashMap<>();
```
bryyeh, line 30: Is this suppose to be deleted?

src/main/java/data/QueryInformation.java
```diff
@@ -27,29 +27,29 @@
   private static final long MILLISECONDS_TO_END_OF_DAY = (24L * 60L * 60L * 1000L) - 1L;
 
   private final TimeZone timeZone;
-  private final ImmutableSet<String> categories;
+  private final ImmutableSet<String> category;
```
acheng1, line 30: nit: if this is a list we should keep plural `categories`

emilyvera: I had it as categories, but in #32 I was told to use category since for the mvp only one category is accepted. Should I still change it back? I could also change it back to categories in the next iteration

PR #40, opened 2020-07-15
Make all receipts display on homepage load

emilyvera:


src/main/webapp/js/index.js
```diff
@@ -12,10 +12,22 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+/** Fetches all receipts from the server and adds them to the DOM. */
+async function displayAllReceipts() {
+  const params = new URLSearchParams();
+  params.append('isNewLoad', 'true');
```
basaam0, line 18: nit: It might be better to use a boolean here if it does not affect the servlet

avagarcia0: I also think it would be better to name this parameter something like getAllReceipts that describes what is actually being requested.

emilyvera: @basaam0 I think params are always sent as a string, so even passing a boolean would require parsing it on the servlet

acheng1: @emilyvera this comment still seems unresolved? +1 to @avagarcia0 about the naming.

emilyvera: @acheng1 I changed it in commit aece6b7 but I think when I merged another branch into this one it reverted back. What's a good workflow to avoid this from happening?

src/test/java/SearchServletTest.java
```diff
@@ -184,6 +184,29 @@ public void queryWithoutStoreAndCategory() throws IOException {
...
+    // id   Timestamp      Price          Store                    Categories
+    // 1  1045237591000    26.12        "walmart"         ["candy", "drink", "personal"]
+    // 2  1560193140000    14.51        "contoso"         ["cappuccino", "sandwich", "lunch"]
+    // 3  1491582960000    29.01   "main st restaurant"   ["food", "meal", "lunch"]
+
+    // Add mock receipts to datastore.
+    TestUtils.addTestReceipts(datastore);
+
+    // Perform doGet - this should retrieve all receipts.
+    when(request.getParameter("isNewLoad")).thenReturn("true");
```
acheng1, line 200: nitnitnit: add a newline after this line (between the arrange part and act)

src/test/java/SearchServletTest.java
```diff
@@ -184,6 +184,29 @@ public void queryWithoutStoreAndCategory() throws IOException {
...
+    servlet.doGet(request, response);
+    writer.flush();
+
+    // Make sure all receipts retrieved by finding the ids in the writer.
```
acheng1, line 204: If someone added another receipt to `addTestReceipts()`, this test would still pass but it would not have checked that all receipts had been returned right? I think I might put all of the receipts (aka BlobKeys) in `addTestReceipts()` in a static list and this test can call `addTestReceipts(datastore, receipts)` with the receipts to add. Then the assert can check receipts match.

src/main/webapp/js/index.js
acheng1: That seems fine but you should have gotten merge conflicts in that case, sounds like you did not?

emilyvera: I didn't get conflicts, it seems like it just overwrote it :/

emilyvera: Oh wait...I just double checked and it's getAllReceipts :)

acheng1: Oh haha. Glad it worked!

src/test/java/SearchServletTest.java
emilyvera: BlobKeys?

acheng1: Sorry I meant `Entity`s (note none of this code is tested and you would need a small Receipt wrapper class):

```
private static final Entity RECEIPT_ENTRY_1 = createEntity(new BlobKey("test"), "img/walmart-receipt.jpg", 26.12, "walmart", ImmutableSet.of("candy", "drink", "personal"), "");
...

private static final ImmutableList<Entity> receiptEntries = ImmutableSet.of(RECEIPT_ENTRY_1, ....);
..
addTestReceipts(datastore, receiptEntries);
```
Then in the test:

```
Receipt[] receipts = new Gson().fromJson(stringWriter.toString(), Receipt[].class);
Assert.assertTrue(receipts.size(), receiptEntries.length());
Assert.assertTrue(receipts.stream().anyMatch(...)); // check id
...
...
```

src/test/java/SearchServletTest.java
```diff
@@ -184,6 +191,37 @@ public void queryWithoutStoreAndCategory() throws IOException {
...
+
+    when(request.getParameter("isNewLoad")).thenReturn("true");
+
+    // Perform doGet - this should retrieve all receipts.
+    servlet.doGet(request, response);
+    writer.flush();
+
+    // Make sure all receipts retrieved by checking their ids.
+    Gson gson =
+        new GsonBuilder().registerTypeAdapterFactory(ImmutableAdapterFactory.forGuava()).create();
```
acheng1, line 214: 👍

PR #43, opened 2020-07-15
Sanitize and format store name stored in receipt entity.

basaam0:
Similar to #41, the store name field in a receipt entity must also be formatted (lowercase, leading and trailing spaces omitted, and only one space separating words) to conform with the search implementation.

### DIFF
--- a/src/test/java/UploadReceiptServletTest.java
+++ b/src/test/java/UploadReceiptServletTest.java
@@ -112,7 +112,7 @@ public final class UploadReceiptServletTest {
   private static final String LABEL = "Label";
   private static final Text RAW_TEXT = new Text("raw text");
   private static final double PRICE = 5.89;
-  private static final String STORE = "McDonald's";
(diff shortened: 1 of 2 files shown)


src/test/java/UploadReceiptServletTest.java
```diff
@@ -243,6 +243,31 @@ public void doPostUploadsReceiptToDatastoreDevServer() throws IOException {
...
+    when(ReceiptAnalysis.serveImageText(new URL(LIVE_SERVER_ABSOLUTE_URL)))
+        .thenReturn(ANALYSIS_RESULTS);
+
+    servlet.doPost(request, response);
+
+    Query query = new Query("Receipt");
+    PreparedQuery results = datastore.prepare(query);
+    Entity receipt = results.asSingleEntity();
+
+    String formattedStore = "trader joe's";
```
bryyeh, line 267: nit: expectedStore

PR #47, opened 2020-07-16
Parse categories from Natural Language API

avagarcia0:
Parse categories like "/Food & Drink/Restaurants" into "Food & Drink" and "Restaurants".

### DIFF
--- a/src/main/java/servlets/ReceiptAnalysis.java
+++ b/src/main/java/servlets/ReceiptAnalysis.java
@@ -44,6 +44,7 @@ import java.net.URL;
 import java.util.Collections;
 import java.util.Set;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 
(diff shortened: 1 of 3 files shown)


avagarcia0:
> Is `master` the correct base branch?

Nope, sorry. That should be fixed now.

src/main/java/servlets/ReceiptAnalysis.java
```diff
@@ -161,10 +162,11 @@ private static String retrieveText(ByteString imageBytes)
 
       ClassifyTextResponse response = client.classifyText(request);
 
-      // TODO: Parse category strings into more natural categories
+      // Parse category strings into more natural categories
+      // e.g. "/Food & Drink/Restaurants" becomes "Food & Drink" and "Restaurants"
       categories = response.getCategoriesList()
                        .stream()
-                       .map(category -> category.getName())
+                       .flatMap(category -> Stream.of(category.getName().substring(1).split("/")))
```
acheng1, line 169: If a computation isn’t self-explanatory, I would consider moving the code to a helper method and use method reference instead.

PR #48, opened 2020-07-17
Dependency fixes

avagarcia0:
* Remove duplicate test dependencies
* Resolve dependency conflict

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -107,50 +107,6 @@
       <version>1.9.80</version>
       <scope>test</scope>
     </dependency>
-
-    <!-- Test Dependencies -->
-    <dependency>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -48,7 +48,13 @@
     <dependency>
       <groupId>com.google.api</groupId>
       <artifactId>gax</artifactId>
-      <version>1.57.1</version>
+      <version>1.56.0</version>
```
acheng1, line 51: We need an older version of this?

avagarcia0: Yeah, the newer version causes a conflict since vision and natural language depend on 1.56.0.

PR #45, opened 2020-07-16
Generate categories using Natural Language API

avagarcia0:

PR #50, opened 2020-07-17
Store categories generated by the Natural Language API.

basaam0:
The set of categories stored for a particular receipt entity is the combination of the categories assigned by the user with the categories generated by feeding the raw text into the Natural Language API (from #45).

### DIFF
--- a/src/test/java/UploadReceiptServletTest.java
+++ b/src/test/java/UploadReceiptServletTest.java
@@ -341,7 +341,8 @@ public final class UploadReceiptServletTest {
   }
 
   @Test
-  public void doPostSanitizesCategories() throws IOException, ReceiptAnalysisException {
+  public void doPostSanitizesUserAndGeneratedCategories()
+      throws IOException, ReceiptAnalysisException {
     helper.setEnvIsLoggedIn(true);
 
     String[] userCategories =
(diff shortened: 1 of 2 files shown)


src/test/java/UploadReceiptServletTest.java
```diff
@@ -105,13 +109,18 @@
   private static final long IMAGE_SIZE_0MB = 0;
   private static final String HASH = "35454B055CC325EA1AF2126E27707052";
 
-  private static final String[] CATEGORIES = new String[] {"burger", "fast food", "restaurant"};
-  private static final Collection<String> CATEGORIES_COLLECTION = Arrays.asList(CATEGORIES);
+  private static final String[] USER_CATEGORIES =
+      new String[] {"burger", "fast food", "restaurant"};
+  private static final Set<String> GENERATED_CATEGORIES = Collections.singleton("dining");
```
avagarcia0, line 114: Prefer using ImmutableSet.

src/test/java/UploadReceiptServletTest.java
```diff
@@ -312,17 +344,19 @@ public void doPostRemovesDuplicateCategories() throws IOException, ReceiptAnalys
   public void doPostSanitizesCategories() throws IOException, ReceiptAnalysisException {
```
acheng1, line 344: nitnit: maybe update the test name to say that it tests that categories are sanitized in addition to returning both user and generated categories?

PR #60, opened 2020-07-20
Add category elements as needed

avagarcia0:
Instead of having three category elements already exist, create new elements as they are needed. This allows categories to be spaced out better.

image


src/main/webapp/js/receipt-analysis.js
```diff
@@ -39,3 +42,14 @@ function getDateFromTimestamp(timestamp) {
   const timeZoneId = new Intl.DateTimeFormat().resolvedOptions().timeZone;
   return new Date(time).toLocaleString('en-US', {timeZone: timeZoneId});
 }
+
+/** Builds the div element for a category along with its children. */
+function buildCategoryElement(category) {
+  const categoryElement = document.createElement('div');
+  categoryElement.className = 'col d-flex justify-content-center';
+  categoryElement.innerHTML =
```
emilyvera, line 50: I like this approach, but I think we should use a template instead and make clones of it as needed

PR #62, opened 2020-07-20
Capitalize store and category names

avagarcia0:
image


src/main/webapp/js/receipt-analysis.js
```diff
@@ -54,3 +56,10 @@ function buildCategoryElement(category) {
 
   return categoryElement;
 }
+
+/** Capitalize the first letter of each word in a string. */
```
basaam0, line 60: nit: "Capitalizes"

src/main/webapp/js/receipt-analysis.js
```diff
@@ -17,7 +17,7 @@ function loadReceiptAnalysis() {
   const parameters = new URLSearchParams(location.search);
 
   const date = getDateFromTimestamp(parameters.get('timestamp'));
-  const storeName = parameters.get('store');
+  const storeName = capitalizeFirstLetters(parameters.get('store'));
```
acheng1, line 20: I think this is okay but will convert names like KFC, McDonald's, Chick-fil-A to Kfc, Mcdonald's and Chick-fil-a right?

avagarcia0: Yeah, it's the same function that's used for the home page.

avagarcia0:
> We don't have any JS tests yet is that right?

That's correct.

src/main/webapp/js/receipt-analysis.js
bryyeh: Instead of copying this function, could you reuse the one in index.js? Or move it to a common file for both files to use.

avagarcia0: Yeah, we were discussing creating a common file for JS functions, so I'll just move it there once we've settled on a filename.

bryyeh: Sounds good. Can you add a TODO comment here for that?

PR #63, opened 2020-07-20
Fix redirection from home page to login page.

basaam0:
Before, a logged-out user would have elements rendered on the home page before being redirected to the login page.
- Hides the home page body by default


src/main/webapp/index.html
```diff
@@ -47,12 +47,14 @@
 <link rel="stylesheet" href="https://unpkg.com/nouislider@10.0.0/distribute/nouislider.min.css" />
 
 <!-- Custom -->
+<link rel="stylesheet" href="css/style.css" />
```
acheng1, line 50: Just curious why the separate style.css and script.js files?  Are there more specific names that could be used?

basaam0: I put some of the code common to multiple pages in the style.css and script.js files. Would a name like "shared.css" be fine?

acheng1: "shared.css" or "common.css" seems fine to me.

PR #65, opened 2020-07-21
Fix redirection from upload page to login page for a logged-out user.

basaam0:
Similar to #63, but for the upload page.
- Hides the upload page body by default


src/main/webapp/css/upload.css
```diff
@@ -1,3 +1,7 @@
+body {
+  display: none;
+}
```
avagarcia0, line 3: This should be added to style.css since it's being applied to every page now. We may also want to create a JavaScript file shared by all pages to avoid repeated code.

basaam0: Shared CSS and JS have been added to #63.

PR #66, opened 2020-07-21
Fix redirection from receipt analysis page to login page for a logged-out user.

basaam0:
Similar to #63, but for the receipt analysis page.
- Hides the receipt analysis page body by default

### DIFF
--- a/src/main/webapp/js/receipt-analysis.js
+++ b/src/main/webapp/js/receipt-analysis.js
@@ -18,7 +18,7 @@
  */
 function load() {
   /* global loadPage */
-  loadPage(loadReceiptAnalysis);
+  loadPage(loadReceiptAnalysis);  // From js/script.js.
 }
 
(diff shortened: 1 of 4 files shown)


src/main/webapp/js/receipt-analysis.js
```diff
@@ -12,6 +12,15 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+/**
+ * Loads the page if the user is logged in. Otherwise, redirects to the
+ * login page.
+ */
+function load() {
+  /* global loadPage */
+  loadPage(loadReceiptAnalysis);
```
bryyeh, line 21: I dislike that it's not clear where the method loadPage() is from. I wonder if we can add an import or a comment?

basaam0: Added a comment. I tried importing, but that would require configuring all the js files as modules in the linter.

PR #67, opened 2020-07-21
Wrap category text to prevent category badges from overlapping

avagarcia0:
image


avagarcia0:
> I like this solution! I handled wrapping in a different way, so we should probably decide which way we like better so our pages are cohesive

I think the best option is to combine our approaches to handle cases like this:

image

My only issue with this is that I'd prefer it to default to my method first, but it defaults to yours instead. I'm guessing there's no easy way to change that, though I don't know Bootstrap CSS well enough to be sure.

PR #68, opened 2020-07-21
Fix categories wrapping on home page

emilyvera:
Same issue as #67 : fix wrapping so long category names don't overlap 

Screenshot 2020-07-21 at 11 22 20 AM


src/main/webapp/js/index.js
```diff
@@ -34,6 +34,8 @@ async function checkAuthentication() {
     window.location.replace('/login.html');
   }
 
+  // TODO: account.email will get the person's email
```
bryyeh, line 37: This looks unrelated to this pull request?

PR #71, opened 2020-07-21
Widen date and price default values

emilyvera:
Previously, the default value for date was the past month and for price was $20-$80. I've changed these to the past 6 months and $0-$80 to increase results if the user doesn't manually set these.


emilyvera:
> Any particular reason you chose $80 as the maximum price?

Nope, it was arbitrary. If you have suggestions I can change it!

PR #74, opened 2020-07-22
Add logo detection to receipt analysis.

basaam0:
- Adds a store property to `AnalysisResults`
- Adds logo detection to the image annotation request to Cloud Vision
- Adds a builder class for `AnalysisResults`


src/main/java/data/AnalysisResults.java
```diff
@@ -36,4 +39,41 @@ public String getRawText() {
   public ImmutableSet<String> getCategories() {
     return categories;
   }
-}
\ No newline at end of file
+
+  public Optional<String> getStore() {
+    return this.store;
```
avagarcia0, line 44: Does `this` need to be here?

src/main/java/servlets/ReceiptAnalysis.java
```diff
@@ -141,15 +145,24 @@ private static String retrieveText(ByteString imageBytes)
...
-      // First element has the entire raw text from the image
-      EntityAnnotation annotation = response.getTextAnnotationsList().get(0);
+      // First element has the entire raw text from the image.
+      EntityAnnotation textAnnotation = response.getTextAnnotationsList().get(0);
 
-      rawText = annotation.getDescription();
+      String rawText = textAnnotation.getDescription();
+      analysisBuilder = new AnalysisResults.Builder(rawText);
+
+      // If a logo was detected with a confidence above the threshold, use it to set the store.
```
emilyvera, line 154: I'm guessing the threshold is out of 1? And the closer to 1 you are the more accurate the result? Maybe add a comment so this is clear and no assumptions need to be made

src/test/java/ReceiptAnalysisTest.java
```diff
@@ -84,6 +87,10 @@
   private static final ImmutableSet<String> CATEGORIES =
       ImmutableSet.of(GENERAL_CATEGORY_NAME, SPECIFIC_CATEGORY_NAME);
 
+  private static final Optional<String> STORE = Optional.of("Google");
+  private static final float LOGO_CONFIDENCE = 0.6f;
```
emilyvera, line 91: Why is the confidence a float here and a double in ReceiptAnalysis file?

src/test/java/ReceiptAnalysisTest.java
```diff
@@ -214,4 +232,42 @@ public void analyzeImageAtThrowsIfTextRequestFails()
...
+        ClassificationCategory.newBuilder().setName(CATEGORY_NAME).build();
+    ClassifyTextResponse classifyResponse =
+        ClassifyTextResponse.newBuilder().addCategories(category).build();
+    when(languageClient.classifyText(any(ClassifyTextRequest.class))).thenReturn(classifyResponse);
+  }
+
+  private ImmutableList<AnnotateImageRequest> createImageRequest() {
+    Image image = Image.newBuilder().setContent(IMAGE_BYTES).build();
+    List<Feature> features =
+        Arrays.asList(Feature.newBuilder().setType(Feature.Type.TEXT_DETECTION).build(),
```
emilyvera, line 262: You could do ImmutableList.copyOf(Feature.newBuilder()...) to make this an ImmutableList instead of a List

src/main/java/servlets/ReceiptAnalysis.java
```diff
@@ -141,15 +145,24 @@ private static String retrieveText(ByteString imageBytes)
             "Received image annotation response without text annotations.");
       }
```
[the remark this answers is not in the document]
basaam0: Do you mean handling a response without a logo detected?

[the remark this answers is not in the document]
avagarcia0: I mean handling a response that has a logo annotation but no text annotations.

src/test/java/ReceiptAnalysisTest.java
```diff
@@ -49,6 +49,8 @@
 import java.io.IOException;
 import java.io.InputStream;
 import java.net.URL;
+import java.util.Arrays;
```
avagarcia0, line 52: I believe this import is no longer necessary.

src/main/java/servlets/ReceiptAnalysis.java
```diff
@@ -42,6 +42,7 @@
 import java.io.IOException;
 import java.io.InputStream;
 import java.net.URL;
+import java.util.Arrays;
```
avagarcia0, line 45: I believe this import is no longer necessary.

src/main/java/servlets/ReceiptAnalysis.java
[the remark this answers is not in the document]
basaam0: Added a TODO

src/test/java/UploadReceiptServletTest.java
```diff
@@ -56,6 +56,7 @@
 import java.util.HashMap;
```
basaam0, line 56: I accidentally merged #75 too early, so most of this file has already been reviewed.

src/main/java/data/AnalysisResults.java
```diff
@@ -36,4 +39,41 @@ public String getRawText() {
...
     return categories;
   }
-}
\ No newline at end of file
+
+  public Optional<String> getStore() {
+    return store;
+  }
+
+  public static class Builder {
```
acheng1, line 47: Nice!

src/main/java/data/AnalysisResults.java
```diff
@@ -23,10 +24,12 @@
 public class AnalysisResults {
   private final String rawText;
   private final ImmutableSet<String> categories;
+  private final Optional<String> store;
 
-  public AnalysisResults(String rawText, Set<String> categories) {
+  public AnalysisResults(String rawText, Set<String> categories, Optional<String> store) {
```
acheng1, line 29: You can make this constructor private now...

src/main/java/data/AnalysisResults.java
```diff
@@ -36,4 +39,41 @@ public String getRawText() {
...
+    public Builder setStore(String store) {
+      this.store = Optional.of(store);
+      return this;
+    }
+
+    public AnalysisResults build() {
+      // No logo was detected.
+      if (store == null) {
+        store = Optional.empty();
+      }
```
acheng1, line 74: nit: you can use `Optional.ofNullable()` - Returns an Optional describing the specified value, if non-null, otherwise returns an empty Optional.

src/main/java/data/AnalysisResults.java
```diff
@@ -36,4 +39,37 @@ public String getRawText() {
...
+      return this;
+    }
+
+    public Builder setStore(String store) {
+      this.store = Optional.of(store);
+      return this;
+    }
+
+    public AnalysisResults build() {
+      store = Optional.ofNullable(store).map(value -> value.get());
```
acheng1, line 71: could you make `setStore()` take an optional directly?

acheng1, line 71: Does `store = Optional.ofNullable(store);` not work?

basaam0: `store` is of type `Optional<String>`, so that would make a `Optional<Optional<String>>`.

basaam0: Should I instead make the `store` property of the Builder a `String`, and then create the `Optional` in the `build()` method using `ofNullable`?

acheng1: I see. So when do we expect that store would be null?

basaam0: For the case where no logo is detected, `setStore()` would not be called and the `AnalysisResults` would have the store set to an empty `Optional`.

acheng1: Oh, can you default it to empty: `private Optional<String> store = Optional.empty();`

PR #78, opened 2020-07-22
Analytics page: store chart UI

emilyvera:
A simple UI for the analytics page 
- Button to go back to homepage
- Chart displaying user spending at each store

Note: data is static for now

Screenshot 2020-07-22 at 6 28 00 PM

### DIFF
--- a/src/main/webapp/analytics.html
+++ b/src/main/webapp/analytics.html
@@ -1,0 +1,10 @@
+<!DOCTYPE html>
+<meta charset="utf-8" />
+<meta name="viewport" content="width=device-width, initial-scale=1" />
+<title>Spending Analytics</title>
+
+<script src="https://www.gstatic.com/charts/loader.js"></script>
+<script src="script.js"></script>
+
+<h1>Analytics</h1>
+<div id="chart-container"></div>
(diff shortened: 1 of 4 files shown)


src/main/webapp/analytics.html
```diff
@@ -0,0 +1,47 @@
...
+  integrity="sha384-OgVRvuATP1z7JjHLkuOU7Xw704+h835Lr+6QL9UvYjZE3Ipu6Tp75j7Bh/kR0JKI"
+  crossorigin="anonymous"
+></script>
+
+<!-- Google charts API -->
+<script src="https://www.gstatic.com/charts/loader.js"></script>
+
+<!-- Custom -->
+<link rel="stylesheet" href="css/analytics.css" />
+<script src="js/analytics.js"></script>
```
avagarcia0, line 37: You should also add the shared CSS and JS files here, and make sure to redirect to the login page when needed. Since we haven't settled on names for the shared files yet, you could probably just add these tasks as TODOs for now.

src/main/webapp/index.html
```diff
@@ -55,18 +55,23 @@
...
-        </a>
+      <div class="col-6">
+        <div class="row justify-content-end">
+          <div class="col-auto mb-2">
+            <a href="/analytics.html" class="btn btn-info btn-sm">Spending analytics</a>
+          </div>
+          <div class="col-auto">
+            <a href="/upload.html" class="btn btn-info btn-sm">Add a receipt</a>
+          </div>
+        </div>
```
basaam0, line 74: Repeated from #73, maybe it should be the base branch

PR #80, opened 2020-07-23
Add input box for editing total price on the receipt analysis page

avagarcia0:
image


src/main/webapp/js/receipt-analysis.js
```diff
@@ -83,3 +86,39 @@ function capitalizeFirstLetters(lowercasedString) {
...
+}
+
+/**
+ * Converts a string value into a number, removing all non-numeric characters.
+ */
+function convertStringToNumber(string) {
+  return Number(String(string).replace(/[^0-9.]+/g, ''));
+}
+
+/**
```
emilyvera, line 106: Some of these methods seem to be repeated from Basaam's page. Are we still making a shared js/css file to handle this?

PR #81, opened 2020-07-23
Add input box for editing date on the receipt analysis page

avagarcia0:
image


src/main/webapp/js/receipt-analysis.js
```diff
@@ -57,14 +57,15 @@ async function updateReceipt(event) {
   // Prevent the default action of reloading the page on form submission.
   event.preventDefault();
 
-  // TODO: Get transaction date.
```
basaam0, line 60: Eventually, the receipt id will also need to be added to the form data for the edit servlet (#83).

PR #83, opened 2020-07-23
Create edit servlet.

basaam0:
Servlet called by receipt analysis page to edit the receipt categories, store name, price, and/or transaction date.


src/main/java/servlets/EditReceiptServlet.java
```diff
@@ -0,0 +1,125 @@
...
+    if (!userService.isUserLoggedIn()) {
+      logger.warning(USER_NOT_LOGGED_IN_WARNING);
+      response.setStatus(HttpServletResponse.SC_FORBIDDEN);
+      response.getWriter().println(USER_NOT_LOGGED_IN_WARNING);
+      return;
+    }
+
+    Entity receipt;
+    try {
+      long id = Long.parseLong(request.getParameter("id"));
```
avagarcia0, line 78: If you want the receipt analysis page to send an ID, you'll have to modify upload.js to send the ID when it redirects to the receipt analysis page.

basaam0: Added ID in #88.

src/main/java/servlets/EditReceiptServlet.java
```diff
@@ -0,0 +1,125 @@
...
+      receipt = createUpdatedReceipt(request, id);
+    } catch (EntityNotFoundException | InvalidPriceException | InvalidDateException
+        | NumberFormatException e) {
```
emilyvera, line 81: Maybe use a better name that just "e"

src/main/java/servlets/EditReceiptServlet.java
```diff
@@ -0,0 +1,125 @@
...
+      logger.warning(e.toString());
+      response.setStatus(HttpServletResponse.SC_BAD_REQUEST);
+      response.getWriter().println(e.toString());
+      return;
+    }
+
+    datastore.put(receipt);
```
emilyvera, line 88: Should catch DatastoreFailureException here

PR #86, opened 2020-07-24
Refactor TestUtils to work with old tests and new incoming ones

emilyvera:


src/test/java/TestUtils.java
```diff
@@ -29,27 +29,47 @@ private TestUtils() {
...
+            createEntity(/* userId = */ "123", /* timestamp = */ 1491582960000L,
+                new BlobKey("test"), "img/restaurant-receipt.jpeg", 29.01, "main street restaurant",
+                ImmutableSet.of("food", "meal", "lunch"), ""));
+
+    return entities;
+  }
+
+  /** Creates and returns a single Receipt entity. */
+  public static Entity createEntity(String userId, long timestamp, BlobKey blobkey, String imageUrl,
+      Double price, String store, ImmutableSet<String> categories, String rawText) {
```
basaam0, line 72: Is there a reason for using the `Double` wrapper class instead of `double` here?

avagarcia0: It looks like a typo given that the createEntity calls all use a `double`.

emilyvera: In #87 there's a test where price is null. The Double wrapper here will allow for this null check

src/test/java/TestUtils.java
```diff
@@ -29,27 +29,47 @@ private TestUtils() {
...
+
+            createEntity(/* userId = */ "123", /* timestamp = */ 1491582960000L,
+                new BlobKey("test"), "img/restaurant-receipt.jpeg", 29.01, "main street restaurant",
+                ImmutableSet.of("food", "meal", "lunch"), ""));
+
+    return entities;
+  }
+
+  /** Creates and returns a single Receipt entity. */
+  public static Entity createEntity(String userId, long timestamp, BlobKey blobkey, String imageUrl,
```
acheng1, line 71: See my comments in https://github.com/googleinterns/step12-2020/pull/87 ...

PR #87, opened 2020-07-24
Spending analytics helper class

emilyvera:
Made a helper class to calculate user spending analytics. This class only computes analytics by store, but may include analytics by categories if we chose to do that next

### DIFF
--- a/src/main/java/data/SpendingAnalytics.java
+++ b/src/main/java/data/SpendingAnalytics.java
@@ -1,0 +1,58 @@
+// Copyright 2019 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.sps.data;
+
+import com.google.appengine.api.datastore.Entity;
+import com.google.common.base.Strings;
+import com.google.common.collect.ImmutableSet;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.stream.Collectors;
+
+/** Class that computes and stores user's spending analytics. */
+public class SpendingAnalytics {
+  private final HashMap<String, Double> storeAnalytics;
+
+  public SpendingAnalytics(ImmutableSet<Entity> allReceipts) {
+    storeAnalytics = new HashMap<>();
+
+    for (Entity receipt : allReceipts) {
+      String store = (String) receipt.getProperty("store");
+      Double price = (Double) receipt.getProperty("price");
+
+      // Don't add to hashmap if either store or price are invalid.
+      if (!Strings.isNullOrEmpty(store) && price != null) {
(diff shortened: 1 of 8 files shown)


src/main/java/data/SpendingAnalytics.java
```diff
@@ -0,0 +1,58 @@
...
+        if (storeAnalytics.containsKey(store)) {
+          storeAnalytics.put(store, storeAnalytics.get(store) + price);
+        } else {
+          storeAnalytics.put(store, price);
+        }
+      }
+    }
+  }
+
+  /** Capitalizes the first letter of each word in a string. */
```
acheng1, line 48: nit: can we handle all of the presentation formatting on the frontend since we're already doing similar formatting (e.g., https://github.com/googleinterns/step12-2020/blob/03275387fd5381fbc8ada073a2cc149f0c2c7192/src/main/webapp/js/index.js#L136)

src/test/java/SpendingAnalyticsTest.java
```diff
@@ -0,0 +1,162 @@
...
+  public void tearDown() {
+    helper.tearDown();
+  }
+
+  @Test
+  public void storeAnalyticsWithUniqueStoreNames() {
+    // Only one receipt for each store added:
+    // Walmart: $26.12, Contoso: $14.51, Main Street Restaurant: $29.01
+
+    ImmutableSet<Entity> receipts = TestUtils.createEntities();
```
acheng1, line 68: I don't see a `createEntities()` method in TestUtils?

src/test/java/SpendingAnalyticsTest.java
```diff
@@ -0,0 +1,162 @@
...
+  @Test
+  public void nullStoreNameNotIncludedInAnalytics() {
+    // null store name should be ignored:
+    // null: $26.12, Walmart: $26.12
+
+    ImmutableSet<Entity> receipts = new ImmutableSet.Builder<Entity>()
+                                        .add(TestUtils.createEntity(USER_ID, TIMESTAMP, BLOB_KEY,
+                                            IMAGE_URL, PRICE, STORE, CATEGORIES, RAW_TEXT))
+                                        .add(TestUtils.createEntity(USER_ID, TIMESTAMP, BLOB_KEY,
+                                            IMAGE_URL, PRICE, /*store=*/null, CATEGORIES, RAW_TEXT))
```
acheng1, line 114: nitnitnit: (add spaces)
`/* store= */ null`

src/test/java/SpendingAnalyticsTest.java
acheng1: Nevermind, found it: https://github.com/googleinterns/step12-2020/pull/86

acheng1, line 68: I would probably move the arrangements in `TestUtils.createEntities()` here. We want to ensure that a test can be understood without looking at the definitions of helper methods and not obscure the relation between the arrangement and assertions.

src/test/java/SpendingAnalyticsTest.java
```diff
@@ -0,0 +1,162 @@
...
+
+    SpendingAnalytics analytics = new SpendingAnalytics(receipts);
+
+    // Check that only one stores is returned and it's the expected one.
+    Assert.assertEquals(1, analytics.getStoreAnalytics().size());
+    Assert.assertTrue(analytics.getStoreAnalytics().containsKey("Walmart"));
+  }
+
+  @Test
+  public void emptyStoreNameNotIncludedInAnalytics() {
```
acheng1, line 125: good test coverage!

PR #89, opened 2020-07-24
Spending analytics servlet

emilyvera:
Add servlet to compute user analytics using SpendingAnalytics helper class. Again, this only computes analytics by store name, but categories may be added later

### DIFF
--- a/src/test/java/SpendingAnalyticsServletTest.java
+++ b/src/test/java/SpendingAnalyticsServletTest.java
@@ -1,0 +1,99 @@
+// Copyright 2019 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.sps;
+
+import static org.mockito.Mockito.when;
+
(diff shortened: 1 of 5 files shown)


src/test/java/SpendingAnalyticsServletTest.java
```diff
@@ -0,0 +1,99 @@
...
+
+  @Test
+  public void doGetWithReceiptsInDatastore() throws IOException {
+    // Receipts in datastore:
+    // Walmart: $26.12, Contoso: $14.51, Main Street Restaurant: $29.01
+    TestUtils.addTestReceipts(datastore);
+
+    // Perform doGet.
+    servlet.doGet(request, response);
+    writer.flush();
```
basaam0, line 77: I don't think this comment is necessary,

src/test/java/SpendingAnalyticsServletTest.java
```diff
@@ -0,0 +1,98 @@
...
+  @After
+  public void tearDown() {
+    helper.tearDown();
+  }
+
+  @Test
+  public void doGetWithReceiptsInDatastore() throws IOException {
+    // Receipts in datastore:
+    // Walmart: $26.12, Contoso: $14.51, Main Street Restaurant: $29.01
+    TestUtils.addTestReceipts(datastore);
```
acheng1, line 73: nit: similar comment about making sure that a test can be understood without looking at the definitions of helper method...

emilyvera: The comments on lines 71/72 are supposed to show what the test is without all the clutter of the setup. Do you think I should still change these tests?

acheng1: Yeah, generally we would prefer code that is self explanatory over adding comments.

emilyvera: thanks that makes sense

PR #90, opened 2020-07-24
Analytics js refactored to use datastore data

emilyvera:
Old analytics.js used static test data. Now it uses data pulled from the datastore from the servlet

### DIFF
--- a/src/main/webapp/js/analytics.js
+++ b/src/main/webapp/js/analytics.js
@@ -12,29 +12,34 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-/** Sets callbacks for chart instance. */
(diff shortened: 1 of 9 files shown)


src/main/webapp/js/analytics.js
```diff
@@ -12,29 +12,34 @@
...
+  const response = await fetch('/compute-analytics');
+  const analytics = await response.json();
+
+  storeData = new google.visualization.DataTable();
+
+  storeData.addColumn('string', 'Store');
+  storeData.addColumn('number', 'Total');
+
+  for (const [store, total] of Object.entries(analytics)) {
+    storeData.addRow([store, total]);
```
acheng1, line 32: See my comment in the other PR, but I'd do the formatting of the store name here.

PR #97, opened 2020-07-27
Add loading image to upload page.

basaam0:
While the receipt upload and analysis occur, the upload page will loop a loading animation before redirecting to the receipt analysis page.
upload


src/main/webapp/css/upload.css
```diff
@@ -50,6 +50,29 @@ small {
   font-size: 12px;
 }
 
+.ldBar {
```
emilyvera, line 53: rename to load-bar?

src/main/webapp/js/upload.js
```diff
@@ -86,6 +92,36 @@ async function fetchBlobstoreUrl() {
...
+  document.body.style.cursor = 'wait';
+
+  const submitButton = document.getElementById('submit-receipt');
+  submitButton.disabled = true;
+  submitButton.innerText = 'Analyzing...';
+
+  // Display the loading image, which is hidden by default.
+  const loadingBar = document.getElementsByClassName('ldBar')[0].ldBar;
+  loadingBar.set(1);
+  document.getElementById('loading').style.display = 'block';
```
emilyvera, line 109: could this styling go in the css file?

src/main/webapp/css/upload.css
basaam0: The ldBar class is required for the loading bar library.

https://loading.io/progress/

src/main/webapp/upload.html
```diff
@@ -36,7 +39,17 @@ <h1>Upload a Receipt</h1>
       <label id="receipt-filename-label" class="custom-file-label" for="receipt-image-input">Choose file</label>
     </div>
     <small class="form-text text-muted">Only JPEG images under 5 MB are accepted.</small>
-    <br />
+
+    <div id="loading" class="hidden">
+      <div
+        class="ldBar"
```
acheng1, line 45: See https://google.github.io/styleguide/htmlcssguide.html#ID_and_Class_Name_Delimiters

Also could we use "loading" instead of abbreviating?

PR #99, opened 2020-07-27
Add categories to spending analytics helper class

emilyvera:
SpendingAnalytics.java now calculates both store and category analytics. Added tests for this too


src/main/java/data/SpendingAnalytics.java
```diff
@@ -17,27 +17,63 @@
...
+
+    // Don't add to hashmap if either store or price are invalid.
+    if (!Strings.isNullOrEmpty(store) && price != null) {
+      // Either update the existing entry or add a new one if it doesn't exist.
+      if (storeAnalytics.containsKey(store)) {
+        storeAnalytics.put(store, storeAnalytics.get(store) + price);
+      } else {
+        storeAnalytics.put(store, price);
+      }
+    }
```
basaam0, line 54: Based on [tott/553](https://drive.google.com/file/d/1iNgKtlzK6FnIUnSa87gfayfzrj56L3eb/view), I think that making the null check a guard clause instead of nesting the if statements would make the code easier to follow.

src/main/java/data/SpendingAnalytics.java
```diff
@@ -17,27 +17,63 @@
...
     for (Entity receipt : allReceipts) {
-      String store = (String) receipt.getProperty("store");
-      Double price = (Double) receipt.getProperty("price");
+      updateStoreAnalytics(receipt, storeAnalytics);
 
-      // Don't add to hashmap if either store or price are invalid.
-      if (!Strings.isNullOrEmpty(store) && price != null) {
-        if (storeAnalytics.containsKey(store)) {
-          storeAnalytics.put(store, storeAnalytics.get(store) + price);
+      updateCategoryAnalytics(receipt, categoryAnalytics);
```
basaam0, line 37: Are `storeAnalytics` and `categoryAnalytics` necessary to include as arguments? They can be accessed as instance variables.

src/test/java/SpendingAnalyticsTest.java
```diff
@@ -158,4 +160,90 @@ public void storeWithNullPriceNotIncludedInAnalytics() {
...
+            .add(TestUtils.createEntity(USER_ID, TIMESTAMP, BLOB_KEY, IMAGE_URL,
+                /* price = */ 14.51, STORE, ImmutableSet.of("cappuccino", "food"), RAW_TEXT))
+            .build();
+
+    SpendingAnalytics analytics = new SpendingAnalytics(receipts);
+
+    Assert.assertEquals(26.12, analytics.getCategoryAnalytics().get("candy"), ERROR_THRESHOLD);
+    Assert.assertEquals(26.12, analytics.getCategoryAnalytics().get("drink"), ERROR_THRESHOLD);
+    Assert.assertEquals(14.51, analytics.getCategoryAnalytics().get("cappuccino"), ERROR_THRESHOLD);
+    Assert.assertEquals(14.51, analytics.getCategoryAnalytics().get("food"), ERROR_THRESHOLD);
```
basaam0, line 182: Based on [tott/538](https://drive.google.com/file/d/1uzaIjN8oLqdVL6Sp0VE7ZnBoehEfg_rz/view), it might be useful to save the price values in local variables.

src/main/java/data/SpendingAnalytics.java
```diff
@@ -17,33 +17,78 @@
...
+    if (price == null) {
+      return;
+    }
+
+    ImmutableSet<String> categories;
+
+    // If categories is null for a particular receipt, just skip it and continue.
+    try {
+      categories = ImmutableSet.copyOf((ArrayList) receipt.getProperty("categories"));
+    } catch (NullPointerException exception) {
```
acheng1, line 73: See Item 69 from Effective Java: Use exceptions only for exceptional conditions 

It would be better to check for `receipt.getProperty("categories") == null`

PR #100, opened 2020-07-27
Categories added to spending analytics servlet

emilyvera:
SpendingAnalyticsServlet now handles both stores and categories

### DIFF
--- a/src/main/java/servlets/SpendingAnalyticsServlet.java
+++ b/src/main/java/servlets/SpendingAnalyticsServlet.java
@@ -19,6 +19,7 @@ import com.google.appengine.api.datastore.DatastoreServiceFactory;
 import com.google.appengine.api.datastore.Entity;
 import com.google.appengine.api.datastore.FetchOptions;
 import com.google.appengine.api.datastore.Query;
+import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
 import com.google.gson.Gson;
 import com.google.sps.data.SpendingAnalytics;
@@ -45,15 +46,15 @@ public class SpendingAnalyticsServlet extends HttpServlet {
 
   @Override
   public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
-    HashMap<String, Double> storeAnalytics = getStoreAnalytics();
+    ImmutableList<HashMap<String, Double>> analytics = getAnalytics();
 
     Gson gson = new Gson();
     response.setContentType("application/json;");
-    response.getWriter().println(gson.toJson(storeAnalytics));
+    response.getWriter().println(gson.toJson(analytics));
   }
 
-  /** Returns analytics HashMap mapping a store to the total amount spent there. */
-  private HashMap<String, Double> getStoreAnalytics() {
+  /** Returns ImmutableList with information for both category and store analytics. */
+  private ImmutableList<HashMap<String, Double>> getAnalytics() {
     Query query = new Query("Receipt");
     ImmutableSet<Entity> allReceipts = datastore.prepare(query)
                                            .asList(FetchOptions.Builder.withDefaults())
@@ -61,6 +62,7 @@ public class SpendingAnalyticsServlet extends HttpServlet {
                                            .collect(ImmutableSet.toImmutableSet());
 
     SpendingAnalytics analytics = new SpendingAnalytics(allReceipts);
-    return analytics.getStoreAnalytics();
+
+    return ImmutableList.of(analytics.getStoreAnalytics(), analytics.getCategoryAnalytics());
   }
 }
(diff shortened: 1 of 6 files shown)


src/main/java/servlets/SpendingAnalyticsServlet.java
```diff
@@ -45,22 +46,23 @@ public SpendingAnalyticsServlet(DatastoreService datastore) {
...
     Query query = new Query("Receipt");
     ImmutableSet<Entity> allReceipts = datastore.prepare(query)
                                            .asList(FetchOptions.Builder.withDefaults())
                                            .stream()
                                            .collect(ImmutableSet.toImmutableSet());
 
     SpendingAnalytics analytics = new SpendingAnalytics(allReceipts);
-    return analytics.getStoreAnalytics();
+
+    return ImmutableList.of(analytics.getStoreAnalytics(), analytics.getCategoryAnalytics());
```
basaam0, line 66: Returning a `SpendingAnalytics` object and converting that to JSON would achieve the same goal as creating an `ImmutableList` out of each hash map.

src/test/java/SpendingAnalyticsServletTest.java
```diff
@@ -76,22 +75,35 @@ public void doGetWithReceiptsInDatastore() throws IOException {
     writer.flush();
 
     // Make sure all stores returned in HashMap.
-    HashMap<String, Double> storeAnalytics =
-        new ObjectMapper().readValue(stringWriter.toString(), HashMap.class);
+    System.out.println(stringWriter.toString());
+    HashMap<String, Double> storeAnalytics = TestUtils.parseStoreAnalytics(stringWriter.toString());
```
basaam0, line 79: Is this print statement meant to be included?

src/test/java/TestUtils.java
```diff
@@ -93,4 +94,18 @@ public static void setSearchServletRequestParameters(HttpServletRequest request,
     when(request.getParameter("min")).thenReturn(minPrice);
     when(request.getParameter("max")).thenReturn(maxPrice);
   }
+
+  /** Parses a string containing store analytics into a hashmap representation. */
+  public static HashMap<String, Double> parseStoreAnalytics(String str) throws IOException {
+    String stores = str.substring(str.indexOf("{"), str.indexOf("}") + 1);
+    stores = stores.substring(stores.lastIndexOf("{"), stores.indexOf("}") + 1);
```
bryyeh, line 101: Is there a more precise way to extract the stores? Perhaps we match for "storeAnalytics"? It's difficult to understand the "{" matching.

src/test/java/SpendingAnalyticsServletTest.java
```diff
@@ -74,35 +86,25 @@ public void doGetWithReceiptsInDatastore() throws IOException {
...
     HashMap<String, Double> categoryAnalytics =
         TestUtils.parseCategoryAnalytics(stringWriter.toString());
-    Assert.assertEquals(4, categoryAnalytics.size());
-    Assert.assertTrue(categoryAnalytics.containsKey("candy"));
-    Assert.assertTrue(categoryAnalytics.containsKey("drink"));
-    Assert.assertTrue(categoryAnalytics.containsKey("cappuccino"));
-    Assert.assertTrue(categoryAnalytics.containsKey("food"));
+
+    Assert.assertTrue(EXPECTED_STORE_ANALYTICS.equals(storeAnalytics));
+    Assert.assertTrue(EXPECTED_CATEGORY_ANALYTICS.equals(categoryAnalytics));
```
bryyeh, line 94: Assert.assertEquals() would be more clear.

src/test/java/SpendingAnalyticsServletTest.java
```diff
@@ -35,6 +35,9 @@
 import org.mockito.MockitoAnnotations;
 
 public final class SpendingAnalyticsServletTest {
+  private static final HashMap<String, Double> EXPECTED_STORE_ANALYTICS = new HashMap<>();
+  private static final HashMap<String, Double> EXPECTED_CATEGORY_ANALYTICS = new HashMap<>();
```
bryyeh, line 39: I recommend initializing the map here so it's only done once for all the tests. I think the below should work:
`new HashMap(ImmutableMap.of("walmart", 26.12, "contoso", 14.51, "main street restaurant", 29.01))`

src/test/java/TestUtils.java
```diff
@@ -95,17 +95,33 @@ public static void setSearchServletRequestParameters(HttpServletRequest request,
...
-    stores = stores.substring(stores.lastIndexOf("{"), stores.indexOf("}") + 1);
-    return new ObjectMapper().readValue(stores, HashMap.class);
+    String key = "storeAnalytics";
+
+    // Set indexes to extract the first {...} from str.
+    int startIndex = str.indexOf(key) + key.length() + 2;
+    int endIndex = str.indexOf("}") + 1;
+
+    String json = str.substring(startIndex, endIndex);
+    return new ObjectMapper().readValue(json, HashMap.class);
```
bryyeh, line 110: Do you think this would be a better way to parse the string?
https://stackoverflow.com/questions/2591098/how-to-parse-json-in-java

PR #101, opened 2020-07-27
Initial date parsing

avagarcia0:
* Split raw text on whitespace and filter out any tokens without digits
* Check for a token that is formatted like a date
* If found, attempt to parse it into a date
* If successful, convert that date into a timestamp


src/main/java/data/AnalysisResults.java
```diff
@@ -67,13 +75,23 @@ public Builder setStore(String store) {
...
       // No logo was detected.
       if (store == null) {
         store = Optional.empty();
       }
 
-      return new AnalysisResults(rawText, categories, store);
+      // No date was detected.
+      if (timestamp == null) {
+        timestamp = Optional.empty();
+      }
```
basaam0, line 92: This check could be avoided by initializing `timestamp` to `Optional.empty()`. I did this for `store` in the updated master branch.

src/main/java/servlets/ReceiptAnalysis.java
```diff
@@ -189,14 +205,91 @@ private static AnalysisResults analyzeImage(ByteString imageBytes)
...
+   */
+  private static void checkForParsableDate(
+      AnalysisResults.Builder analysisBuilder, Stream<String> tokens) {
+    Stream<String> dates = tokens.filter(ReceiptAnalysis::isDate);
+    // Assume that the first date on the receipt is the transaction date.
+    Optional<String> firstDate = dates.findFirst();
+
+    if (firstDate.isPresent()) {
+      addDateIfValid(analysisBuilder, firstDate.get());
+    }
```
basaam0, line 228: `Optional.ifPresent()` could be used here.

https://docs.oracle.com/javase/8/docs/api/java/util/Optional.html#ifPresent-java.util.function.Consumer-

src/main/java/servlets/ReceiptAnalysis.java
```diff
@@ -56,6 +63,11 @@
   // minimum confidence score that a detected logo must have to be considered significant for
   // receipt analysis.
   private static final float LOGO_DETECTION_CONFIDENCE_THRESHOLD = 0.4f;
+  // Matches strings containing at least one digit.
+  private static final Pattern digitRegex = Pattern.compile(".*\\d.*");
+  // Matches strings in U.S. date format.
+  private static final Pattern dateRegex =
+      Pattern.compile("\\d?\\d([/-])\\d?\\d\\1\\d{2}(\\d{2})?");
```
basaam0, line 70: Eventually, a more specific regex should be used here because the month and day of a date are restricted to ranges, and this would match an invalid date like 13/32/2020.

src/main/java/servlets/ReceiptAnalysis.java
```diff
@@ -189,14 +205,89 @@ private static AnalysisResults analyzeImage(ByteString imageBytes)
...
+      // Invalid month or day
+      return;
+    }
+  }
+
+  /**
+   * DateTimeFormatter assumes that two-digit years are in the 2000s. This method checks if the
+   * given date is after the current date and if so, subtracts 100 years to move it to the 1900s.
+   */
+  private static ZonedDateTime fixYearIfInFuture(ZonedDateTime dateAndTime) {
```
emilyvera, line 262: hopefully no one tries to upload a receipt that old either way lol! but good edge case

src/test/java/ReceiptAnalysisTest.java
```diff
@@ -89,6 +90,17 @@
   private static final float LOGO_CONFIDENCE = 0.6f;
   private static final float LOGO_CONFIDENCE_BELOW_THRESHOLD = 0.2f;
 
+  private static final String RAW_TEXT_WITH_DATE = "the date is 05-08-2020";
```
acheng1, line 93: Maybe add a test for '/' separator?

PR #102, opened 2020-07-27
Add categories to charts ui

emilyvera:
- Add categories column chart to the UI
- Add an "other" pie slice to handle cases where there are too many stores to reasonably fit on the chart
- Categories chart limited to top 10 categories

Screenshot 2020-07-28 at 4 30 45 PM

### DIFF
--- a/src/main/webapp/js/analytics.js
+++ b/src/main/webapp/js/analytics.js
@@ -40,7 +40,7 @@ function handleStoreChart(analytics) {
   storeData.addColumn('string', 'Store');
   storeData.addColumn('number', 'Total');
 
-  for (const [store, total] of Object.entries(analytics[0])) {
+  for (const [store, total] of Object.entries(analytics.storeAnalytics)) {
     storeData.addRow([capitalizeFirstLetters(store), total]);
   }
 
@@ -57,7 +57,7 @@ function handleCategoryChart(analytics) {
   categoryData.addColumn('string', 'Category');
   categoryData.addColumn('number', 'Total');
 
-  for (const [store, total] of Object.entries(analytics[1])) {
+  for (const [store, total] of Object.entries(analytics.categoryAnalytics)) {
     categoryData.addRow([capitalizeFirstLetters(store), total]);
   }
 
--- a/src/main/webapp/js/analytics.js
+++ b/src/main/webapp/js/analytics.js
@@ -26,8 +26,6 @@ async function computeChartAnalytics() {
   const response = await fetch('/compute-analytics');
   const analytics = await response.json();
 
-  console.log(analytics);
-
   handleStoreChart(analytics);
   handleCategoryChart(analytics);
 }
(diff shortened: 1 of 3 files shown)


src/main/webapp/js/analytics.js
```diff
@@ -16,51 +16,106 @@
...
+  const chartHeight =
+      document.getElementById('category-chart').getBoundingClientRect().height;
+
+  const options = {
+    title: 'Spending habits by category',
+    pieHole: 0.3,
+    width: chartWidth,
+    legend: {alignment: 'center'},
+    chartArea: {width: chartWidth, height: chartHeight * 0.6},
+    sliceVisibilityThreshold: .03,
```
basaam0, line 109: I'm not sure that using a pie chart would accurately convey category data. Wouldn't having 1 receipt with 2 categories show 50% spent on each? This can be further discussed, but out of the [Chart Gallery](https://developers.google.com/chart/interactive/docs/gallery) I think either the column chart or bar chart would better represent spending.

avagarcia0: I agree with this. If we *really* wanted to use a pie chart here, we could pick some category (say, the first one) for each receipt to exclusively fall under, but I feel like the selection would end up being arbitrary.

src/main/webapp/js/analytics.js
```diff
@@ -16,51 +16,116 @@
...
+let categoryData;
 
-/** Computes and populates DataTable with analytics for stores chart. */
-async function computeStoreChartsAnalytics() {
+/** Computes and populates DataTable with analytics for both charts. */
+async function computeChartAnalytics() {
   const response = await fetch('/compute-analytics');
   const analytics = await response.json();
 
+  console.log(analytics);
```
basaam0, line 29: Extra print statement.

src/main/webapp/js/analytics.js
```diff
@@ -16,51 +16,116 @@
...
+ * Populates DataTable with category data and passes that to draw method.
+ * @param {Array} analytics First entry is store data, second is category.
+ */
+function handleCategoryChart(analytics) {
+  categoryData = new google.visualization.DataTable();
+
+  categoryData.addColumn('string', 'Category');
+  categoryData.addColumn('number', 'Total');
+
+  for (const [store, total] of Object.entries(analytics.categoryAnalytics)) {
```
bryyeh, line 62: Is this suppose to be "category" instead of "store"? Same on the next line.

src/main/webapp/js/analytics.js
```diff
@@ -16,51 +16,116 @@
...
 
+  console.log(analytics);
+
+  handleStoreChart(analytics);
+  handleCategoryChart(analytics);
+}
+
+/**
+ * Populates DataTable with store data and passes that to draw method.
+ * @param {Array} analytics First entry is store data, second is category.
```
bryyeh, line 37: This is not an array right? I believe it's json. Comment will probably also need to be changed to reflect that.

PR #104, opened 2020-07-27
Update upload page instructions.

basaam0:
upload


basaam0:
> What's the difference between these two instructions?

They're pretty much the same, I just reworded it slightly so it was different from the heading.

PR #105, opened 2020-07-28
Initial price parsing

avagarcia0:
* Check for tokens formatted as prices
* If at least one price is found, keep the largest price
* Add the kept price to the AnalysisResults


src/main/java/servlets/ReceiptAnalysis.java
```diff
@@ -288,6 +290,45 @@ private static boolean isDate(String token) {
...
+   * Checks the raw text in the builder for prices that can be parsed. The largest price found, if
+   * it exists, is added to the builder.
+   */
+  private static void checkForParsablePrices(AnalysisResults.Builder analysisBuilder) {
+    Stream<String> prices =
+        getTokensFromRawText(analysisBuilder.getRawText()).filter(ReceiptAnalysis::isPrice);
+    // Assume that the largest price on the receipt is the total price.
+    double largestPrice = prices.mapToDouble(ReceiptAnalysis::parsePrice)
+                              .reduce(Double.NEGATIVE_INFINITY, Double::max);
+
```
basaam0, line 303: The `Stream.max()` method can be used for this special case of reduction.

https://docs.oracle.com/javase/8/docs/api/java/util/stream/Stream.html#max-java.util.Comparator-

src/main/java/servlets/ReceiptAnalysis.java
```diff
@@ -214,12 +216,12 @@ private static AnalysisResults analyzeImage(ByteString imageBytes)
...
-   * builder as a timestamp.
+   * Checks the raw text in the builder for a date that can be parsed. If one is found, it is added
+   * to the builder as a timestamp.
    */
-  private static void checkForParsableDate(
-      AnalysisResults.Builder analysisBuilder, Stream<String> tokens) {
-    Stream<String> dates = tokens.filter(ReceiptAnalysis::isDate);
+  private static void checkForParsableDate(AnalysisResults.Builder analysisBuilder) {
+    Stream<String> dates =
+        getTokensFromRawText(analysisBuilder.getRawText()).filter(ReceiptAnalysis::isDate);
```
basaam0, line 224: Is there a benefit in filtering for `hasDigit` since the `isDate` and `isPrice` filters are applied anyways?

src/main/java/servlets/ReceiptAnalysis.java
avagarcia0: The issue with that option is that I'd have to check for both an empty `Optional` and `Double.NEGATIVE_INFINITY`, hence why I used reduce with an identity.

src/main/java/data/AnalysisResults.java
```diff
@@ -26,13 +26,15 @@
   private final ImmutableSet<String> categories;
   private final Optional<String> store;
   private final Optional<Long> timestamp;
+  private final Optional<Double> price;
 
-  public AnalysisResults(
-      String rawText, Set<String> categories, Optional<String> store, Optional<Long> timestamp) {
+  public AnalysisResults(String rawText, Set<String> categories, Optional<String> store,
+      Optional<Long> timestamp, Optional<Double> price) {
```
bryyeh, line 32: Instead of "timestamp", can this be "purchaseTimestamp" or "transactionTimestamp". "timestamp" is unclear by itself. Change the other uses in this class to match.

src/main/java/data/AnalysisResults.java
```diff
@@ -51,11 +53,16 @@ public String getRawText() {
...
   }
 
+  public Optional<Double> getPrice() {
+    return price;
+  }
+
   public static class Builder {
     private final String rawText;
     private ImmutableSet<String> categories;
     private Optional<String> store;
```
bryyeh, line 63: Should this also be initialized with Optional.empty()

src/main/java/servlets/ReceiptAnalysis.java
```diff
@@ -127,8 +127,8 @@ private static AnalysisResults analyzeImage(ByteString imageBytes)
     AnalysisResults.Builder analysisBuilder = retrieveText(imageBytes);
     ImmutableSet<String> categories = categorizeText(analysisBuilder.getRawText());
 
-    Stream<String> tokens = getTokensFromRawText(analysisBuilder.getRawText());
-    checkForParsableDate(analysisBuilder, tokens);
+    checkForParsableDate(analysisBuilder);
```
bryyeh, line 130: I feel it's more readable if you pass in the raw text into checkForParsableDate() and have it return the timestamp. Same on the line below.

src/main/java/servlets/ReceiptAnalysis.java
```diff
@@ -267,25 +267,56 @@ private static ZonedDateTime fixYearIfInFuture(ZonedDateTime dateAndTime) {
...
+   * it exists, is added to the builder.
    */
-  private static boolean isDate(String token) {
-    return dateRegex.matcher(token).matches();
+  private static void checkForParsablePrices(AnalysisResults.Builder analysisBuilder) {
+    // Assume that the largest price on the receipt is the total price.
+    double largestPrice = getTokensFromRawText(analysisBuilder.getRawText())
+                              .filter(ReceiptAnalysis::isPrice)
+                              .mapToDouble(ReceiptAnalysis::parsePrice)
+                              .reduce(Double.NEGATIVE_INFINITY, Double::max);
```
bryyeh, line 292: Instead of using Double.NEGATIVE_INFINITY to represent a price that can't be parsed, I would use Optional. You can have parsePrice return Optional<Double>, and then filter out empty optionals.

src/test/java/ReceiptAnalysisTest.java
```diff
@@ -221,6 +233,48 @@ public void analyzeImageAtUrlReturnsDateIn1900s() throws IOException, ReceiptAna
     Assert.assertEquals(TIMESTAMP_IN_1900S, results.getTimestamp());
   }
 
+  @Test
+  public void analyzeImageAtUrlReturnsPrice() throws IOException, ReceiptAnalysisException {
```
bryyeh, line 237: Can you rename the tests so they have the format methodBeingTested_situationBeingTested_expectedResult().

Example: analyzeImageAtUrl_rawTextWithPrice_returnsPrice()

src/test/java/ReceiptAnalysisTest.java
```diff
@@ -102,6 +102,16 @@
   private static final Optional<Long> TIMESTAMP_IN_1900S =
       Optional.of(Long.valueOf(INSTANT_IN_1900S.toEpochMilli()));
 
+  private static final String RAW_TEXT_WITH_PRICE = "the price is $12.77 in total";
+  private static final String RAW_TEXT_WITH_PRICE_NO_DOLLAR_SIGN = "the price is 12.77 in total";
+  private static final String RAW_TEXT_WITH_MULTIPLE_PRICES =
+      "the items cost $8.99, $2.79, and $1.99, so the total is $12.77 after the $1.00 discount";
```
bryyeh, line 108: I'm am of the opinion that if it's used once in a test, you should create it in the test. This way the list of global variables doesn't get too long to read.

src/test/java/ReceiptAnalysisTest.java
```diff
@@ -102,6 +102,16 @@
   private static final Optional<Long> TIMESTAMP_IN_1900S =
       Optional.of(Long.valueOf(INSTANT_IN_1900S.toEpochMilli()));
 
+  private static final String RAW_TEXT_WITH_PRICE = "the price is $12.77 in total";
```
bryyeh, line 105: Have the string include the variable PRICE. This way it's easy to see that the asserted price is in the raw text.

Example: "the price is $" + PRICE + " in total".

src/main/java/data/AnalysisResults.java
avagarcia0: Yeah, that'll be changed once I merge master in.

src/main/java/servlets/ReceiptAnalysis.java
avagarcia0: I wrote it this way to avoid the extra complexity involved in returning Optionals. Since it's an private helper method, I don't think the interface is all that important, hence why I focused on simplicity instead.

src/main/java/servlets/ReceiptAnalysis.java
avagarcia0: Again, I avoided Optionals because I preferred the simpler method rather than adding in an extra filter and map. In this case, I also chose to use an identity instead of Optionals because that's how I learned to use reduce in ML, so it seemed more natural.

src/test/java/ReceiptAnalysisTest.java
avagarcia0: I talked with Tony about this and he said that the current names seem to include the required information. Thus, I don't want to spend time switching the format of every test method name unless it's really necessary.

avagarcia0: I've updated the new tests to use this format. The rest should be handled in #124.

PR #109, opened 2020-07-28
Integrate transaction date parsing with upload servlet.

basaam0:
Replaces the user-assigned timestamp with the parsed timestamp (#101) for creating a new receipt.


basaam0:
> Looks like this doesn't compile?

This PR depends on #101 and the latest version of master, so I'm waiting for that to merge.

src/test/java/UploadReceiptServletTest.java
```diff
@@ -394,43 +393,31 @@ public void doPostThrowsIfUserIsLoggedOut() throws IOException {
   }
 
   @Test
-  public void doPostThrowsIfDateIsInTheFuture() throws IOException, ReceiptAnalysisException {
+  public void doPostUploadsReceiptWithoutTimestampIfParsedDateIsInTheFuture()
```
bryyeh, line 396: These test names are getting kinda long. We usually like our test names to be in the format: methodBeingTested_situationBeingTested_expectedResult(). Either in this PR or a separate PR, I would refactor the test names.

Example:
transferMoney_deductsFromSource
transferMoney_addsToDestination
transferMoney_insufficientFunds_throwsException
multiply_twoNegativeValues_returnsPositiveValue
multiply_anythingWithNan_returnsNan
inputIsValid_conversionSucceeds

PR #116, opened 2020-07-30
Add warning for incomplete receipt.

basaam0:
- Fixes issue with fields showing up as "Undefined" on receipt analysis page
- Sets the default value for the date input to the current date
- Removes fields that were not detected from the receipt analysis page query string
- Ensures that all fields have been set before the user can leave the receipt analysis page

### DIFF
--- a/src/main/webapp/js/upload.js
+++ b/src/main/webapp/js/upload.js
@@ -80,24 +80,7 @@ async function uploadReceipt(event) {
   }
 
   const json = (await response.json());
-  const receipt = json.propertyMap;
-  const params = new URLSearchParams();
-  params.append('id', json.key.id);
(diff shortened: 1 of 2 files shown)


src/main/webapp/js/upload.js
```diff
@@ -83,11 +83,21 @@ async function uploadReceipt(event) {
...
+  }
+  if (receipt.price) {
+    params.append('price', receipt.price);
+  }
+  if (receipt.store) {
+    params.append('store', receipt.store);
+  }
+  if (receipt.timestamp) {
+    params.append('timestamp', receipt.timestamp);
+  }
```
bryyeh, line 100: nitnit: it might make sense to pull this into a separate helper method.

PR #118, opened 2020-07-30
Add functionality to save changes button on receipt analysis page.

basaam0:
- Sends a request to the edit servlet when the "Save Changes" button on the edit form is clicked
- Gets receipt entity ID from the query string

### DIFF
--- a/src/main/webapp/js/receipt-analysis.js
+++ b/src/main/webapp/js/receipt-analysis.js
@@ -154,6 +154,9 @@ async function updateReceipt(event) {
   params.append('store', editedReceipt.store);
   params.append('timestamp', editedReceipt.timestamp);
 
+  // Remove warning for unsaved changes.
+  window.onbeforeunload = null;
+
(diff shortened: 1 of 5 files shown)


src/main/webapp/js/receipt-analysis.js
```diff
@@ -123,32 +123,57 @@ async function updateReceipt(event) {
...
+  const json = await response.json();
+  const editedReceipt = json.propertyMap;
+
+  const params = new URLSearchParams();
+  params.append('id', json.key.id);
+  params.append('image-url', editedReceipt.imageUrl.value);
+  params.append('categories', editedReceipt.categories);
+  params.append('price', editedReceipt.price);
+  params.append('store', editedReceipt.store);
+  params.append('timestamp', editedReceipt.timestamp);
```
bryyeh, line 155: nitnit: I think it'd be nice to put this in a helper method.

PR #119, opened 2020-07-30
Add loading cursor after submitting edit form.

basaam0:
Disables the "Save Changes" and "Return to Home" buttons and changes to the loading cursor while the request to the edit servlet is processing.

### DIFF
--- a/src/main/webapp/js/receipt-analysis.js
+++ b/src/main/webapp/js/receipt-analysis.js
@@ -125,7 +125,8 @@ async function updateReceipt(event) {
   document.body.style.cursor = 'wait';
   const saveButton = document.getElementById('submit-receipt');
   const homeButton = document.getElementById('return-home');
-  saveButton.disabled = homeButton.disabled = true;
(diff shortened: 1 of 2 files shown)


src/main/webapp/js/receipt-analysis.js
```diff
@@ -122,6 +122,11 @@ async function updateReceipt(event) {
   // Prevent the default action of reloading the page on form submission.
   event.preventDefault();
 
+  document.body.style.cursor = 'wait';
+  const saveButton = document.getElementById('submit-receipt');
+  const homeButton = document.getElementById('return-home');
+  saveButton.disabled = homeButton.disabled = true;
```
emilyvera, line 128: Not sure if this is relevant since this isn't technically a declaration, but maybe this line should be separated. I also just think it would be better for readability

saveButton.disabled = true;
homeButton.disabled = true;

https://google.github.io/styleguide/jsguide.html#features-one-variable-per-declaration

PR #120, opened 2020-07-30
Improved price parsing

avagarcia0:
Adds a heuristic that checks for a price after the word total. If this heuristic doesn't find anything, then the old heuristic is used.


src/main/java/servlets/ReceiptAnalysis.java
```diff
@@ -281,19 +283,65 @@ private static boolean isDate(String token) {
...
+          && parsePrice(token) != Double.NEGATIVE_INFINITY) {
+        price = parsePrice(token);
+        priceFound = true;
+        priceFoundAfterMostRecentTotal = true;
+      }
+    }
+
+    if (priceFound) {
+      analysisBuilder.setPrice(price);
+      return (true);
```
basaam0, line 323: extra parenthesis?

src/main/java/servlets/ReceiptAnalysis.java
```diff
@@ -281,19 +283,65 @@ private static boolean isDate(String token) {
...
+
+    double price = 0;
+    boolean priceFound = false;
+    boolean priceFoundAfterMostRecentTotal = true;
+
+    // Try to find the first price after each token containing "total", and keep the last one found.
+    for (String token : relevantTokens) {
+      if (containsTotal(token)) {
+        priceFoundAfterMostRecentTotal = false;
+      } else if (isPrice(token) && !priceFoundAfterMostRecentTotal
```
basaam0, line 313: The `isPrice()` check will never be false since the tokens are already filtered so it does not need to be included.

src/main/java/servlets/ReceiptAnalysis.java
```diff
@@ -281,19 +283,65 @@ private static boolean isDate(String token) {
...
+          && parsePrice(token) != Double.NEGATIVE_INFINITY) {
```
basaam0, line 314: Under what conditions would a NumberFormatException occur? Since the token matches the price regex, isn't it guaranteed to be in the correct format?

avagarcia0: If I wrote the price regex correctly, then yes. That is there in case I didn't write it correctly.

src/main/java/servlets/ReceiptAnalysis.java
```diff
@@ -281,19 +283,64 @@ private static boolean isDate(String token) {
...
   private static void checkForParsablePrices(AnalysisResults.Builder analysisBuilder) {
-    // Assume that the largest price on the receipt is the total price.
+    if (!findPriceAfterTotal(analysisBuilder)) {
+      findLargestPrice(analysisBuilder);
+    }
+  }
+
+  /**
+   * Checks the raw text in the builder for a price appearing after the word "total" and adds it to
+   * the builder if found. The return value indicates whether a price was added.
```
emilyvera, line 297: Use a java tag here instead of a sentence. So it would be "@return whether a price was added." 

https://www.oracle.com/technical-resources/articles/java/javadoc-tool.html#format

src/main/java/servlets/ReceiptAnalysis.java
```diff
@@ -281,19 +283,64 @@ private static boolean isDate(String token) {
...
+    ImmutableList<String> relevantTokens =
+        getTokensFromRawText(analysisBuilder.getRawText())
+            .filter(token -> isPrice(token) || containsTotal(token))
+            .collect(ImmutableList.toImmutableList());
+
+    double price = 0;
+    boolean priceFound = false;
+    boolean priceFoundAfterMostRecentTotal = true;
+
+    // Try to find the first price after each token containing "total", and keep the last one found.
```
emilyvera, line 309: If you check the Macy's receipt I uploaded to the shared drive, this is an example of a receipt where it says "Total: 21.00" and further down it says "Your total savings is $6.25." Would this return $6.25? Maybe you can edit this heuristic or add another one to do a combination of highest price and searching for the word "total"

src/test/java/ReceiptAnalysisTest.java
```diff
@@ -275,6 +276,70 @@ public void analyzeImageAt_dateAndPrice_returnsTimestampAndPrice()
...
+    stubAnnotationResponse(LOGO_CONFIDENCE, rawTextWithPriceMixedCaseTotal);
+    stubTextClassification();
+
+    AnalysisResults results = ReceiptAnalysis.analyzeImageAt(url);
+
+    Assert.assertEquals(PRICE, results.getPrice());
+  }
+
+  @Test
+  public void analyzeImageAt_priceAfterTotalIsNotLargest_returnsPriceAfterTotal()
```
emilyvera, line 331: Is this named correctly? I'm not understanding this test case.

emilyvera: I think I understand it now, maybe it could use a clearer name? Not sure if it would be confusing to others as well or if it's just me!

avagarcia0: I'm not sure how to make this clearer. Do you have any suggestions?

src/main/java/servlets/ReceiptAnalysis.java
avagarcia0: As we discussed earlier, combining the two would fail in a case like "Your total before discounts are applied is $21.00, and your total after discounts is $17.75". I think it's best to stick with the current heuristics for the moment.

src/main/java/servlets/ReceiptAnalysis.java
```diff
@@ -123,6 +125,7 @@ private static AnalysisResults analyzeImage(ByteString imageBytes) throws IOExce
 
     // Generate categories and parse date and price if text was extracted.
     if (analysisBuilder.getRawText().isPresent()) {
+      System.out.println(analysisBuilder.getRawText().get());
```
acheng1, line 128: Could we use a logger as was done here: https://github.com/googleinterns/receipt-roundup/blob/b51c880fa819564c35e534761f8a6f811807f47a/src/main/java/servlets/EditReceiptServlet.java#L48

src/main/java/servlets/ReceiptAnalysis.java
```diff
@@ -280,19 +283,66 @@ private static boolean isDate(String token) {
...
-   * it exists, is added to the builder.
+   * Searches the raw text in the builder for the total price using multiple heuristics. If any find
+   * a price, it is added to the builder.
    */
   private static void checkForParsablePrices(AnalysisResults.Builder analysisBuilder) {
-    // Assume that the largest price on the receipt is the total price.
+    if (!findPriceAfterTotal(analysisBuilder)) {
+      findLargestPrice(analysisBuilder);
+    }
+  }
```
acheng1, line 293: I guess I could see this getting very nested 

```
if (!a) {
  if (!b) {
    if (!c) {
      d();
    }
  }
}
```

So how about structuring it as 
```
if (a()) {
  return;
}

if (b) {
  return;
} 
```

src/main/java/servlets/ReceiptAnalysis.java
```diff
@@ -280,19 +283,66 @@ private static boolean isDate(String token) {
...
+    double price = 0;
+    boolean priceFound = false;
+    boolean priceFoundAfterMostRecentTotal = true;
+
+    // Try to find the first price after each token containing "total", and keep the last one found.
+    for (String token : relevantTokens) {
+      if (containsTotal(token)) {
+        priceFoundAfterMostRecentTotal = false;
+      } else if (!priceFoundAfterMostRecentTotal && parsePrice(token) != Double.NEGATIVE_INFINITY) {
+        price = parsePrice(token);
```
acheng1, line 316: nit: this ends up having to call parse price twice. Can we save the result?

src/main/java/servlets/ReceiptAnalysis.java
```diff
@@ -318,6 +368,13 @@ private static boolean isPrice(String token) {
     return priceRegex.matcher(token).matches();
   }
 
+  /**
+   * Checks if the token contains the word "total".
+   */
+  private static boolean containsTotal(String token) {
```
bryyeh, line 374: nit: containsTotalText would be more clear.

src/main/java/servlets/ReceiptAnalysis.java
```diff
@@ -280,19 +283,66 @@ private static boolean isDate(String token) {
...
+   */
+  private static boolean findPriceAfterTotal(AnalysisResults.Builder analysisBuilder) {
+    ImmutableList<String> relevantTokens =
+        getTokensFromRawText(analysisBuilder.getRawText().get())
+            .filter(token -> isPrice(token) || containsTotal(token))
+            .collect(ImmutableList.toImmutableList());
+
+    double price = 0;
+    boolean priceFound = false;
+    boolean priceFoundAfterMostRecentTotal = true;
```
bryyeh, line 309: If you can, it's better to not maintain state over a loop to simplify the code. In this case, I recommend finding all the indexes of the "total" text. Then, get the price after each index.

acheng1: Alternatively, could we use regex to help find all "total" and price pairs and then we could just grab the last pair found?

src/main/java/servlets/ReceiptAnalysis.java
avagarcia0: Yeah, this was just a debugging statement that accidentally got merged in.

src/main/java/servlets/ReceiptAnalysis.java
bryyeh: Hey Ava, I know in our 1:1 we decided it was okay to leave this alone since we were low on time. After discussing with Tony, we decided we would like you to fix this. It would look better for the hiring committee.

PR #121, opened 2020-07-30
Handle receipt analysis errors.

basaam0:
Previously, any error that occurred during the receipt analysis, such as with extracting raw text, detecting the store name, or generating categories, would prevent the user from uploading a receipt.
- Makes the raw text field optional for the `AnalysisResults` class
- Instead of throwing receipt analysis exceptions, return `AnalysisResults` with empty fields, so that the user can manually input them (ex. If no raw text can be extracted from an image, upload a receipt with empty raw text and category fields)

### DIFF
--- a/src/main/java/servlets/ReceiptAnalysisServlet.java
+++ b/src/main/java/servlets/ReceiptAnalysisServlet.java
@@ -16,7 +16,6 @@ package com.google.sps.servlets;
 
 import com.google.gson.Gson;
 import com.google.sps.data.AnalysisResults;
-import com.google.sps.servlets.ReceiptAnalysis.ReceiptAnalysisException;
 import java.io.IOException;
 import java.net.URL;
 import javax.servlet.annotation.WebServlet;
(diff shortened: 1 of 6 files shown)


src/main/java/servlets/ReceiptAnalysisServlet.java
```diff
@@ -41,13 +40,7 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
       return;
     }
 
-    try {
-      results = ReceiptAnalysis.analyzeImageAt(new URL(url));
-    } catch (ReceiptAnalysisException e) {
-      response.setStatus(HttpServletResponse.SC_INTERNAL_SERVER_ERROR);
```
acheng1, line 47: I suppose we could have just returned an empty object here instead of an error?

basaam0: The `ReceiptAnalysisServlet` is not used in the flow, so that would not affect the application.

PR #124, opened 2020-07-30
Refactor names for upload, edit, and analysis unit tests.

basaam0:
Uses methodBeingTested_situationBeingTested_expectedResult() format.

### DIFF
--- a/src/test/java/EditReceiptServletTest.java
+++ b/src/test/java/EditReceiptServletTest.java
@@ -135,7 +135,7 @@ public final class EditReceiptServletTest {
   }
 
   @Test
-  public void doPostEditsReceipt() throws IOException {
+  public void doPost_editsReceipt() throws IOException {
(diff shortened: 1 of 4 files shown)


src/test/java/EditReceiptServletTest.java
```diff
@@ -135,7 +135,7 @@ public void tearDown() {
   }
 
   @Test
-  public void doPostEditsReceipt() throws IOException {
+  public void doPost_editsReceipt() throws IOException {
```
avagarcia0, line 138: Are you allowed to omit the situation being tested?

basaam0: From #109:
> Example:
> transferMoney_deductsFromSource
> transferMoney_addsToDestination
> transferMoney_insufficientFunds_throwsException
> multiply_twoNegativeValues_returnsPositiveValue
> multiply_anythingWithNan_returnsNan
> inputIsValid_conversionSucceeds

I assume yes because some of the examples omit it.

PR #123, opened 2020-07-30
Add userid to search

emilyvera:
Add userId to search query so when you make a search, you only get receipts back that you uploaded and not all receipts in the datastore. 

Also note: when I ran the lint checker it fixed the formatting in 3 random files, but there were no actual code changes there

PR #125, opened 2020-07-31
Add userid to analytics

emilyvera:
Spending analytics now uses userId to calculate analytics. Previously, it was taking everything from the datastore and calcualating based off that, but we want to make sure only the logged in user's receipts are being taken into account.

Note: same as #123 for some reason when I ran the lint checker it fixed the formatting in 3 random files, but there were no actual code changes there


src/main/java/servlets/SpendingAnalyticsServlet.java
```diff
@@ -33,18 +36,30 @@
 /** Servlet that computes user spending analytics from datastore data. */
 @WebServlet("/compute-analytics")
 public class SpendingAnalyticsServlet extends HttpServlet {
+  private static final String AUTHENTICATION_ERROR_MESSAGE =
+      "No Authentication: User must be logged in to search receipts.";
+
   private final DatastoreService datastore;
+  private final UserService userService;
```
basaam0, line 43: `userService` can be initialized here since both constructors create the same object.

PR #127, opened 2020-07-31
Refactor search servlet to handle pagination

emilyvera:
Search servlet handles both forward and backward pagination with 10 receipts displayed per page.


src/main/java/servlets/SearchServlet.java
```diff
@@ -94,17 +106,25 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
+   * @return the input string parsed to a boolean.
+   */
+  private Boolean checkParameter(HttpServletRequest request, String parameter) {
+    return Boolean.parseBoolean(request.getParameter(parameter));
+  }
+
   /** Creates a {@link QueryInformation} based on request parameters. */
-  private QueryInformation createQueryInformation(
-      HttpServletRequest request, HttpServletResponse response)
+  private void createQueryInformation(HttpServletRequest request, HttpServletResponse response)
```
basaam0, line 127: I don't see the response parameter used here.

src/main/java/servlets/SearchServlet.java
```diff
@@ -72,14 +80,18 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
+    ImmutableList<Receipt> receipts = null;
 
-    if (Boolean.parseBoolean(request.getParameter("isNewLoad"))) {
-      receipts = getAllReceipts();
-    } else {
+    if (!checkParameter(request, "isNewSearch") && checkParameter(request, "getNextPage")) {
+      receipts = getNextPageOfReceipts();
+    } else if (!checkParameter(request, "isNewSearch")
+        && checkParameter(request, "getPreviousPage")) {
+      receipts = getPreviousPageOfReceipts();
```
basaam0, line 89: Since `getNextPageOfReceipts()` and `getPreviousPageOfReceipts()` have a very similar structure, I think the code would be more clear if they were combined into one function with a boolean parameter that determines whether to set `startCursor()` or `endCursor()`. Also, maybe the `getNextPage` and `getPreviousPage` request parameters could be combined into one. I do see how this might make things more confusing though, so this is more of an option.

src/test/java/SearchServletTest.java
```diff
@@ -288,4 +305,73 @@ public void checkAuthenticationErrorIsReturned() throws IOException {
...
+  @Test
+  public void paginationNextPage() throws IOException {
+    // This test simulates starting on page 1, then moving to page 2.
+
+    // Add 12 mock receipts to datastore.
+    ImmutableSet<Entity> expectedReceipts = TestUtils.addManyTestReceipts(datastore);
+
+    // Perform doGet - this should retrieve max receipts for page, which is 10.
+    when(request.getParameter("isPageLoad")).thenReturn("true");
+    servlet.doGet(request, response);
```
basaam0, line 318: There should be vertical space between L317 and L318 to separate the "arrange" statements from the "act" statements.

src/test/java/TestUtils.java
```diff
@@ -110,10 +130,26 @@ public static void setSearchServletRequestParameters(HttpServletRequest request,
...
+  /**
+   * Checks receipt ids in the returned page all match the ids in the expected page.
+   * @return true if all expectedPage receipt ids are found in returnedPage, else false.
+   */
+  public static boolean checkIdsMatch(ImmutableList<Entity> expectedPage, Receipt[] returnedPage) {
+    for (Entity expectedEntity : expectedPage) {
+      if (!Arrays.stream(returnedPage)
+               .anyMatch(receipt -> receipt.getId() == expectedEntity.getKey().getId())) {
+        return false;
+      }
```
basaam0, line 151: The stream could also be mapped to the IDs before calling `anyMatch()`.

emilyvera: Can you explain what you mean by that?

basaam0:
```
Stream<Long> ids = Arrays.stream(returnedPage).map(Receipt::getId);
for (...) {
  if (!ids.anyMatch(id -> id == expectedEntity.getKey().getId()))
```

src/main/java/servlets/SearchServlet.java
```diff
@@ -94,17 +106,25 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
     response.setContentType("application/json;");
     response.getWriter().println(gson.toJson(receipts));
   }
 
+  /**
+   * Check if the passed in parameter was true or false.
+   * @param parameter A string representation of a boolean: "true" or "false".
+   * @return the input string parsed to a boolean.
+   */
+  private Boolean checkParameter(HttpServletRequest request, String parameter) {
```
avagarcia0, line 122: The return type should be `boolean` rather than `Boolean`.

src/test/java/TestUtils.java
```diff
@@ -61,6 +66,22 @@ public static Entity addTestReceipt(DatastoreService datastore, String userId, l
     return entities;
   }
 
+  /** Adds 12 receipts to datastore. */
+  public static ImmutableSet<Entity> addManyTestReceipts(DatastoreService datastore) {
```
acheng1, line 70: nit: maybe make the number of receipts a parameter?

src/test/java/TestUtils.java
```diff
@@ -110,10 +132,27 @@ public static void setSearchServletRequestParameters(HttpServletRequest request,
...
   }
+
+  /**
+   * Checks receipt ids in the returned page all match the ids in the expected page.
+   * @return true if all expectedPage receipt ids are found in returnedPage, else false.
+   */
+  public static boolean checkIdsMatch(ImmutableList<Entity> expectedPage, Receipt[] returnedPage) {
+    Supplier<Stream<Long>> ids = () -> Arrays.stream(returnedPage).map(Receipt::getId);
+
+    for (Entity expectedEntity : expectedPage) {
```
acheng1, line 151: Maybe also check the size matches so that returned doesn't contain more than expected?

emilyvera: I assert for that in lines 341 and 378 in SearchServletTest. So by the time this method is called the sizes are guaranteed to be the same already.

acheng1: Would it make sense to put that check in this method since they are tightly coupled (checkIdsMatch() can't validate all id's match without ensuring the size is the same)?

src/main/java/servlets/SearchServlet.java
```diff
@@ -52,16 +58,18 @@
       "No Authentication: User must be logged in to search receipts.";
 
   private final DatastoreService datastore;
-  private final UserService userService;
+  private final UserService userService = UserServiceFactory.getUserService();
+
+  private Cursor cursor;
```
acheng1, line 63: Were you able to test that pagination works locally? If the cursor information is not passed in as part of the request how does getPage() work?

emilyvera: It works both on deployment and locally. The global cursor keeps track of which page its on

acheng1: See https://wiki.sei.cmu.edu/confluence/display/java/MSC11-J.+Do+not+let+session+information+leak+within+a+servlet

Generally not a good idea to use member fields like this as it could end up leaking data from one user/client to another which is why I suggested that you pass the cursor token back to the client and have the client pass it in as a parameter.

index.yaml
```diff
@@ -2,16 +2,36 @@ indexes:
```
acheng1, line 2: I think you put this in a separate pull request?

src/main/java/data/ServerResponse.java
```diff
@@ -0,0 +1,36 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.sps.data;
+
+import com.google.common.collect.ImmutableList;
+
+/** Class that computes and stores user's spending analytics. */
+public class ServerResponse {
```
acheng1, line 20: Maybe consider a more specific name? ServerResponse sounds very generic given that this is only for the spending analytics servlet.

src/main/java/data/ServerResponse.java
```diff
@@ -0,0 +1,36 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.sps.data;
+
+import com.google.common.collect.ImmutableList;
+
+/** Class that computes and stores user's spending analytics. */
```
acheng1, line 19: nit: this class doesn't actually compute anything right? It seems to be a container class for storing the response of the spending analysis servlet?

src/main/java/servlets/SearchServlet.java
```diff
@@ -80,10 +80,10 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
       return;
     }
 
-    ImmutableList<Receipt> receipts = null;
+    ServerResponse serverResponse = null;
 
     if (checkParameter(request, "isPageLoad")) {
-      receipts = getMatchingReceipts(true);
+      serverResponse = getMatchingReceipts(true);
```
acheng1, line 86: I probably should have mentioned this earlier but `getMatchingReceipts(true)` reads as true, this should get matching receipts or `getMatchingReceipts(false)` this should not return matching receipts :). I think ideally I would create separate method names but for now if you just comment the parameter `getMatchingReceipts(/* isPageLoad= */ true)` it would make this a little clearer.

I have a similar comment about `getPage(true)` and `getPage(false)` where I'd probably have methods `getNextPage()` and `getPreviousPage()` instead, but if you could just comment those parameters too that would help.

src/main/java/servlets/SearchServlet.java
```diff
@@ -139,21 +139,26 @@ private void createQueryInformation(HttpServletRequest request)
...
-  private ImmutableList<Receipt> getPage(boolean isNextPage) {
+  private ServerResponse getPage(boolean isNextPage, String encodedCursor) {
+    Cursor cursor = Cursor.fromWebSafeString(encodedCursor);
     FetchOptions options = isNextPage ? FetchOptions.Builder.withStartCursor(cursor)
                                       : FetchOptions.Builder.withEndCursor(cursor);
     options.limit(RECEIPTS_PER_PAGE);
 
     QueryResultList<Entity> results = preparedQuery.asQueryResultList(options);
+    ImmutableList<Receipt> receipts = entitiesListToReceiptsList(results);
+    encodedCursor = results.getCursor().toWebSafeString();
```
acheng1, line 152: Nice, thanks for fixing!

index.yaml
emilyvera: This PR has updated indexes from that one

src/main/java/servlets/SearchServlet.java
```diff
@@ -52,16 +59,17 @@
       "No Authentication: User must be logged in to search receipts.";
 
   private final DatastoreService datastore;
-  private final UserService userService;
+  private final UserService userService = UserServiceFactory.getUserService();
+
+  private PreparedQuery preparedQuery;
+  private QueryInformation queryInformation;
```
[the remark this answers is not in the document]
emilyvera: Yeah QueryInformation can easily be made local but PreparedQuery is needed for pagination so it knows which query to apply the cursor to: https://cloud.google.com/appengine/docs/standard/java/javadoc/com/google/appengine/api/datastore/Cursor. 

I would need to somehow send that information to the client and back like I did with the cursor, but I think it would take a lot of restructuring :/ I can try to do that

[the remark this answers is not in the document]
acheng1: When the user clicks on next or previous, can you just pass in all of the input parameters again used in `createQueryInformation()` to recreate the PreparedQuery?

src/main/java/data/SearchServletResponse.java
```diff
@@ -16,7 +16,7 @@
 
 import com.google.common.collect.ImmutableList;
 
-/** Wrapper class for a search servlet doGet request. */
+/** Wrapper class for infromation sent on a search servlet doGet request. */
```
acheng1, line 19: nit: typo - information

also "doGet response"

src/main/java/servlets/SearchServlet.java
```diff
@@ -132,70 +139,65 @@ private void createQueryInformation(HttpServletRequest request)
...
+  /**
+   * Gets first receipts page from a new query.
+   * @return list of receipts as entities.
+   */
+  private QueryResultList<Entity> getFirstPage(Query query) {
+    PreparedQuery preparedQuery = datastore.prepare(query);
+    QueryResultList<Entity> results =
+        preparedQuery.asQueryResultList(FetchOptions.Builder.withLimit(RECEIPTS_PER_PAGE));
+
+    return preparedQuery.asQueryResultList(FetchOptions.Builder.withLimit(RECEIPTS_PER_PAGE));
```
acheng1, line 178: `return results` ?

src/main/webapp/js/index.js
```diff
@@ -38,35 +39,31 @@ async function loadLogoutButton(account) {
...
   searchReceipts(params);
 }
 
 /** Fetches next receipts page from the server and adds it to the DOM. */
 function getNextPageOfReceipts() {
   const params = new URLSearchParams();
-  params.append('isPageLoad', 'false');
-  params.append('isNewSearch', 'false');
+  params.append('isPageLoad', isPageLoad);
+  params.append('isNewSearch', !isPageLoad);
```
acheng1, line 63: Seems like you can just eliminate isNewSearch from the javascript and servlet altogether?

src/main/java/servlets/SearchServlet.java
```diff
@@ -72,14 +77,15 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
+    Query query = null;
+    QueryInformation queryInformation = null;
 
-    if (Boolean.parseBoolean(request.getParameter("isNewLoad"))) {
-      receipts = getAllReceipts();
-    } else {
+    // Query is set differently based on type of search.
+    if (checkParameter(request, "isPageLoad")) {
+      query = getQuery(/* isPageLoad = */ true, queryInformation);
+    } else if (checkParameter(request, "isNewSearch")) {
```
acheng1, line 86: This can just be `} else {`?

PR #128, opened 2020-07-31
Pagination UI

emilyvera:
- Add back and forth buttons to flip through receipt pages
- Add their function handlers to the js file.
- Add a scroll back to top feature when next or back button pressed.

Screenshot 2020-07-31 at 4 22 52 PM

### DIFF
--- a/src/main/webapp/index.html
+++ b/src/main/webapp/index.html
@@ -120,12 +120,10 @@
     <div id="receipts-display" class="row"></div>
     <div class="row">
       <div class="col-6 d-flex justify-content-end pb-4">
-        <button type="button" class="btn btn-info" onclick="scrollToTop(); getPreviousPageOfReceipts()">
-          &lt; Back
-        </button>
+        <button type="button" class="btn btn-info" onclick="getPreviousPageOfReceipts()">&lt; Back</button>
       </div>
       <div class="col-6 d-flex justify-content-start pb-4">
-        <button type="button" class="btn btn-info" onclick="scrollToTop(); getNextPageOfReceipts()">Next &gt;</button>
+        <button type="button" class="btn btn-info" onclick="getNextPageOfReceipts()">Next &gt;</button>
       </div>
     </div>
   </div>
--- a/src/main/webapp/index.html
+++ b/src/main/webapp/index.html
@@ -110,7 +110,7 @@
           </div>
         </div>
         <div class="col-md-1">
-          <button type="submit" class="btn btn-outline-success my-2 my-sm-0" onclick="searchReceipts()">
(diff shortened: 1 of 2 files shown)


src/main/webapp/index.html
```diff
@@ -110,14 +110,24 @@ <h1 class="text-center">Home</h1>
...
     <!-- The columns will automatically stack on top of each other when the screen is less than 768px wide. -->
     <div id="receipts-display" class="row"></div>
+    <div class="row">
+      <div class="col-6 d-flex justify-content-end pb-4">
+        <button type="button" class="btn btn-info" onclick="scrollToTop(); getPreviousPageOfReceipts()">
+          &lt; Back
+        </button>
+      </div>
+      <div class="col-6 d-flex justify-content-start pb-4">
+        <button type="button" class="btn btn-info" onclick="scrollToTop(); getNextPageOfReceipts()">Next &gt;</button>
```
basaam0, line 128: I think it would be better to move the `scrollToTop()` call inside the `getPreviousPageOfReceipts()` and `getNextPageOfReceipts()` methods so that less implementation details are in the HTML.

src/main/webapp/js/index.js
```diff
@@ -57,13 +56,45 @@ async function searchReceipts() {
...
+ */
+async function searchReceipts(params) {
   const response = await fetch(`/search-receipts?${params.toString()}`);
   const receipts = await response.json();
 
   clearExistingDisplay();
   displayReceipts(receipts);
 }
 
+/** Scrolls to the top of the page when called. */
```
basaam0, line 92: nit: "when called" is implicit so I don't think it's necessary to include.

src/main/webapp/js/index.js
```diff
@@ -14,9 +14,7 @@
 
 /* global capitalizeFirstLetters, loadPage */
 
-/**
```
acheng1, line 17: Was this auto formatted?

acheng1, line 17: Was this automatically formatted?

src/main/webapp/js/index.js
```diff
@@ -32,22 +30,23 @@ async function loadLogoutButton(account) {
       `You are signed in as ${account.email}`;
 }
 
-/** Fetches all receipts from the server and adds them to the DOM. */
-async function getAllReceipts() {
+/**
+ * Fetches all receipts from the server and adds them to the DOM.
+ * This will only be called on page load.
+ */
+function getAllReceipts() {
```
bryyeh, line 37: nit: With pagination, it's no longer "all" receipts right? I would rename this and edit the comment.

PR #131, opened 2020-08-03
Update upload page UI.

basaam0:
- Uses outline style for "Add Receipt" button and decreases its size
- Changes background color
- Makes the loading image have a thinner outline with a different color

up1
up2

### DIFF
--- a/src/main/webapp/css/upload.css
+++ b/src/main/webapp/css/upload.css
@@ -8,8 +8,8 @@ h1 {
   text-align: center;
 }
 
-label {
-  font-size: 20px;
+.card {
+  max-width: 600px;
 }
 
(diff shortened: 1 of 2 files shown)


src/main/webapp/css/upload.css
```diff
@@ -18,7 +18,7 @@ input {
 }
 
 #upload-form {
-  background: linear-gradient(#e3ffe7 0%, #d9e7ff 100%);
+  background: #f7f7f7;
```
acheng1, line 21: BTW, are we using this rounded background outline across the other pages as well?

basaam0: Added the outline background from the receipt analysis page.

PR #136, opened 2020-08-04
Add image to home page button.

basaam0:
upload
analysis
analytics


src/main/webapp/js/analytics.js
```diff
@@ -130,3 +130,11 @@ $(window).resize(function() {
   drawStoreChart(storeData);
   drawCategoryChart(categoryData);
 });
+
+/**
+ * Redirects the user to the home page when the home button is
```
emilyvera, line 135: Don't think "when the home button is clicked" is necessary

src/main/webapp/js/receipt-analysis.js
```diff
@@ -242,7 +242,7 @@ function formatCurrency(event) {
 }
 
 /**
- * Redirects the user to the home page when the "Return to Home" button is
+ * Redirects the user to the home page when the home button is
```
emilyvera, line 245: Also edit comment here

src/main/webapp/receipt-analysis.html
```diff
@@ -23,6 +23,10 @@
 <title>Receipt Analysis</title>
 
 <body onload="load()">
+  <button id="home-button" href="/" class="btn btn-info btn-sm mt-2 ml-2" type="button" onclick="redirectHome()">
```
bryyeh, line 26: There is a redirectHome() in common.js and receipt-analysis.js. They're both loaded for this page. I feel that could cause some confusion, maybe name them differently.

src/main/webapp/receipt-analysis.html
```diff
@@ -23,6 +23,16 @@
 <title>Receipt Analysis</title>
 
 <body onload="load()">
+  <button
+    id="home-button"
+    href="/"
```
acheng1, line 28: Why is there an `href` here but not in upload.html?

basaam0: Removed.

PR #137, opened 2020-08-04
Prevent "subtotal" from counting as a total token

avagarcia0:
This should help handle cases like the Bed Bath & Beyond receipt that have "subtotal" but not "total".

### DIFF
--- a/src/main/java/servlets/ReceiptAnalysis.java
+++ b/src/main/java/servlets/ReceiptAnalysis.java
@@ -68,6 +68,8 @@ public class ReceiptAnalysis {
   private static final Pattern priceRegex = Pattern.compile("\\$?\\d+\\.\\d\\d");
   // Matches strings containing the word "total" with any capitalization.
   private static final Pattern totalRegex = Pattern.compile(".*total.*", Pattern.CASE_INSENSITIVE);
+  // Matches strings containing the word "sub" with any capitalization.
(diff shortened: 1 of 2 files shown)


src/main/java/servlets/ReceiptAnalysis.java
```diff
@@ -369,10 +371,10 @@ private static boolean isPrice(String token) {
   }
 
   /**
-   * Checks if the token contains the word "total".
+   * Checks if the token contains "total" and doesn't contain "sub".
    */
   private static boolean containsTotal(String token) {
```
emilyvera, line 376: Renaming this to containsTotalAndNotSub would be more accurate now

src/main/java/servlets/ReceiptAnalysis.java
```diff
@@ -68,6 +68,8 @@
   private static final Pattern priceRegex = Pattern.compile("\\$?\\d+\\.\\d\\d");
   // Matches strings containing the word "total" with any capitalization.
   private static final Pattern totalRegex = Pattern.compile(".*total.*", Pattern.CASE_INSENSITIVE);
+  // Matches strings containing the word "sub" with any capitalization.
+  private static final Pattern subRegex = Pattern.compile(".*sub.*", Pattern.CASE_INSENSITIVE);
```
bryyeh, line 72: Would it be simpler to exactly match for "total"?

PR #140, opened 2020-08-04
More homepage ui changes

emilyvera:
- Make pills and edit/delete buttons teal.
- Add icons to edit/delete buttons
- Make background of search bar clear with gray outline.

Screenshot 2020-08-04 at 5 03 38 PM

Screenshot 2020-08-04 at 5 03 56 PM

### DIFF
--- a/src/main/webapp/css/home.css
+++ b/src/main/webapp/css/home.css
@@ -18,7 +18,7 @@
 }
 
 .search-bar {
-  background-color: #f7f7f7;
+  outline: 2px solid #ededed;
   padding: 0.3% 0 0.8%;
 }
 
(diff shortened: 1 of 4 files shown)


src/main/webapp/css/home.css
```diff
@@ -18,7 +18,7 @@
 }
 
 .search-bar {
-  background-color: #f7f7f7;
+  outline: 2px solid #ededed;
```
basaam0, line 21: Is there a reason not to use the "card" class in Bootstrap to get the gray outline?

emilyvera: Yeah I tried using that class but it messed up the formatting of the components inside

PR #141, opened 2020-08-04
Update login page UI.

basaam0:
login


src/main/webapp/css/login.css
```diff
@@ -1,16 +1,11 @@
 body {
-  background: azure;
+  background: url('/img/landscape.jpg') center/cover no-repeat fixed;
```
emilyvera, line 2: Remove quotes in url

https://google.github.io/styleguide/htmlcssguide.html#CSS_Quotation_Marks

PR #142, opened 2020-08-04
Account for variable numbers of categories

emilyvera:
Previously, there was a static amount of category tags on each receipt card (3). Now it displays any number of categories.

Also, accidentally added a fix to this PR that displays the transaction date in the corrected timezone.

Screenshot 2020-08-04 at 6 36 40 PM
Screenshot 2020-08-04 at 6 36 46 PM


src/main/webapp/js/index.js
```diff
@@ -122,10 +122,13 @@ function createReceiptCardElement(receipt) {
...
+      receiptCardClone.getElementById('categories-container');
+
   const categories = Array.from(receipt.categories);
-  for (let i = 0; i < categories.length && i < 3; i++) {
-    receiptCardClone.querySelector('#c' + i).innerText =
-        capitalizeFirstLetters(categories[i]);
+  for (let i = 0; i < categories.length; i++) {
+    const category = capitalizeFirstLetters(categories[i]);
+    categoriesContainer.appendChild(createCategoryElement(category));
   }
```
basaam0, line 132: Using forEach() would improve readability.

src/main/webapp/js/index.js
```diff
@@ -116,16 +116,19 @@ function createReceiptCardElement(receipt) {
 
   // Fill in template fields with correct information.
   const date = new Date(receipt.timestamp);
-  receiptCardClone.querySelector('#timestamp').innerText = date.toDateString();
+  receiptCardClone.querySelector('#timestamp').innerText =
+      date.toUTCString().substring(0, 16);
```
acheng1, line 120: nit: this isn't related to categories right? Since it's a small change I'd just mention it in the PR description that you are fixing the date issue too.

PR #144, opened 2020-08-08
Update README.

basaam0:
Includes changes from the README doc by Ava, Emily, and I.


acheng1:
Could we remove the gifs with real receipts in them?

basaam0:
> Could we remove the gifs with real receipts in them?

Done.

